PR #28, opened 2017-05-16
Fix segfaults and deregister warnings

### LINKED ISSUE #21: Wingpanel crashes when updating the indicators
Looks like wingpanel is catching the `.so.dpkg-new` & `.so.dpkg-tmp` during an instalation, and by the time they get added the file could be already done. The plugin manager should only load .so's

```[03:09:54.895631 Critical] IndicatorManager.vala:162: /usr/lib/x86_64-linux-gnu/wingpanel/libnetwork.so.dpkg-new: cannot open shared object file: No such file or directory
[03:09:54.895702 Critical] wingpanel_indicator_manager_deregister_indicator: assertion 'indicator != NULL' failed
[03:09:54.929234 Critical] IndicatorManager.vala:162: /usr/lib/x86_64-linux-gnu/wingpanel/libnetwork.so.dpkg-tmp: cannot open shared object file: No such file or directory
[03:09:54.929258 Critical] wingpanel_indicator_manager_deregister_indicator: assertion 'indicator != NULL' failed
warning: Could not load shared library symbols for /usr/lib/x86_64-linux-gnu/wingpanel/libnotifications-indicator.so.dpkg-new.
Do you need "set solib-search-path" or "set sysroot"?
[03:09:56.356550 GLib-GObject-Warning] cannot register existing type 'Indicator'
[03:09:56.356560 GLib-Critical] g_once_init_leave: assertion 'result != 0' failed
[03:09:56.356581 GLib-GObject-Critical] g_object_new: assertion 'G_TYPE_IS_OBJECT (object_type)' failed
[03:09:56.356587 Critical] wingpanel_indicator_set_visible: assertion 'self != NULL' failed
Can't read data for section '.eh_frame' in file '/usr/lib/x86_64-linux-gnu/wingpanel/libnotifications-indicator.so'

Thread 1 "wingpanel" received signal SIGSEGV, Segmentation fault.
```

donadigo:
A classic example of removing something from the list in the iteration loop was causing segfaults because of accessing memory that's non existent. I've also fixed the problem that the indicators were tried to be deregistered even if they weren't registered, resulting in warnings like `[03:09:54.929258 Critical] wingpanel_indicator_manager_deregister_indicator: assertion 'indicator != NULL' failed`. This should partially fix #21. The wingpanel will still restart because there are some indicators that don't put their `Wingpanel.Indicator` in any namespace and that also results in a crash e.g: notifications-indicator. A really easy way of testing that is just duplicating the .so file in the Wingpanel's plugin directory.

### DIFF
--- a/lib/IndicatorManager.vala
+++ b/lib/IndicatorManager.vala
@@ -71,7 +71,7 @@ public class Wingpanel.IndicatorManager : GLib.Object {
     [CCode (has_target = false)]
     private delegate Wingpanel.Indicator? RegisterPluginFunction (Module module, ServerType server_type);
 
-    private Gee.HashMap<string, Wingpanel.Indicator>? indicators = null;
+    private Gee.HashMap<string, Wingpanel.Indicator> indicators;
 
     private FileMonitor? monitor = null;
 
@@ -129,7 +129,10 @@ public class Wingpanel.IndicatorManager : GLib.Object {
                      */
                     load (plugin_path);
                 } else if (event == FileMonitorEvent.DELETED) {
-                    deregister_indicator (plugin_path, indicators.@get (plugin_path));
+                    var indicator = indicators[plugin_path];
+                    if (indicator != null) {
+                        deregister_indicator (plugin_path, indicator);
+                    }
                 }
             });
         } catch (Error error) {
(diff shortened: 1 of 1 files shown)


lib/IndicatorManager.vala
```diff
@@ -313,14 +316,21 @@ public class Wingpanel.IndicatorManager : GLib.Object {
     public void register_indicator (string path, Wingpanel.Indicator indicator) {
         debug ("%s registered", indicator.code_name);
 
+        var deregister_map = new Gee.HashMap<string,Wingpanel.Indicator> ();
```
Philip-Scott: Just a nitpick, but you should add a space after the `,` :smile: 

Other than that, great job. It does fix the warnings i was getting

donadigo:
@Philip-Scott done.

PR #30, opened 2017-05-18
Cycle indicators

### LINKED ISSUE #7: ctrl + tab should cycle indicators
We should be able to cycle through indicators with ctrl + tab (the tab switch shortcut)

Launchpad Details: [#LP1536423](https://bugs.launchpad.net/bugs/1536423) Daniel Fore - 2016-01-21 01:07:03 +0000

donadigo:
Fixes #7. Makes it possible to use Ctrl+Tab to cycle through indicators. Wingpanel will cycle only when some indicator is currently opened, otherwise, never. To determine the next sibling of the currently opened indicator, I've created `get_next_sibling` helper method in `Wingpanel.Panel` which implements a simple algorithm:

1. Check in what menubar the current indicator is located.
2. If on the left:
- if indicator not found in the menubar, return `null`
- if it's the last indicator in this menubar, return the first one from the center menubar (if exists)
- otherwise, return the next one in this menubar

The same goes for the center and right menubar, except that these will return the next menubar so for center menubar the next will be the right one, and for the right menubar (that is the last), the left will be returned. 

The alghoritm is implemented in a way that if in the future we decide to allow developers to put indicators on the left or center, this will still work, as it's not hardcoded to a particular indicators.

I would like someone to look at the algorithm first and see if it's implemented & works correctly to make sure that this doesn't break anything.

Known issues: Slingshot now cannot switch through categories because it has the same shortcut as cycling.

### DIFF
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -65,6 +65,12 @@ public class Wingpanel.PanelWindow : Gtk.Window {
         panel = new Widgets.Panel (popover_manager);
         panel.realize.connect (on_realize);
 
+        var cycle_action = new SimpleAction ("cycle", null);
+        cycle_action.activate.connect (panel.cycle);
+
+        app.add_action (cycle_action);
+        app.add_accelerator ("<Control>Tab", "app.cycle", null);
+
         this.add (panel);
     }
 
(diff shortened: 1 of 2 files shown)


donadigo:
@danrabbit done, you can now cycle back with Ctrl+Shift+Tab. Regarding focusing the indicator with keyboard: I noticed that e.g: Slingshot is fully working and you can start navigating with keyboard in it while cycling, other indicators don't seem to do anything, the wingpanel always focuses the current indicator, so I think this is more of an issue with indicators, not with the wingpanel.

PR #44, opened 2017-07-23
Use the right mutter version

tintou:
Fixes Juno build


tintou:
For the record: the .vapi files and the CMake logic are taken from Gala

PR #45, opened 2017-07-24
Container.vala: use GObject style

danrabbit:
* Use GObject-style
* Reassign ambiguous "wingpanel developers" to elementary LLC


tintou:
Here the C API remains compatible but the Vala API has a tiny incompatibility. I'm not opposed to this change but we should look carefully at the users of the Wingpanel API

danrabbit:
Yeah, to be honest I'm not sure that this particular class is used directly. It seems like this is only intended as a superclass for the actual wingpanel widgets. It should probably be private :man_shrugging:

PR #48, opened 2017-08-29
Workaround Vala API change

tintou:


ricotz:
This would do the trick though.

-    public Gtk.Grid content_widget { get; construct; }
+    public Gtk.Grid content_widget { owned get; construct; }

PR #63, opened 2017-11-12
Calculate actual popover position for mouse event detection

### LINKED ISSUE #62: There's an invisible wall that doesn't let me close all wingpanel indicators when I click on that part of the screen.
I think that clicking in any part of the screen outside the indicator should close it. But the actual behavior is a bit different: it does close automatically when you click in some parts, but it doesn't do anything when you click in some other parts.  

I've made [a video](https://gfycat.com/gifs/detail/SpottedScaredArthropods) explaining this issue. In that video I try to show where is that invisible wall and how it changes depending on the height of the indicator. I also have this [edited screenshot](https://imgur.com/a/Wh6Kb) that shows more or less the same. NOTE: the actual sizes of that invisible wall are probably different, the picture is just an approximation.

Some other things I've found:

* It happens with all indicators, even the Applications indicator.  

* In the video you can see a weird black glitch, maybe is related to this issue.

* It always happens, doesn't matter if you are in the desktop or using an app.

* With the applications indicator, the invisible wall is barely noticeable, but it's still there on the space between the indicator and the screen border.

davidmhewitt:
Fixes #62 

The allocation of the `Gtk.Box` containing the indicator popover is used to determine if a click was inside the popover or not when deciding whether or not that click event should close the currently open popover.

Because an `Allocation` is relative to its parent this was always `(0, 0)` causing there to be a box the size of the current indicator popover to be in the top left that prevented you from clicking to close indicators. So, this PR does some maths to move the allocation calculation underneath the actual popover relative to the mouse co-ordinates.

I considered removing the mouse bounds check altogether as items inside the popover still receive click events properly. But removing it causes the application menu to be closed when clicking a blank space inside it, so not ideal.

### DIFF
--- a/src/Services/PopoverManager.vala
+++ b/src/Services/PopoverManager.vala
@@ -102,6 +102,27 @@ public class Wingpanel.Services.PopoverManager : Object {
             Gtk.Allocation allocation;
             popover.get_allocation (out allocation);
 
+            Gtk.Allocation indicator_allocation;
+            current_indicator.get_allocation (out indicator_allocation);
+
+            Gtk.Allocation container_allocation;
+            current_indicator.get_parent ().get_allocation (out container_allocation);
+
+            int wingpanel_width;
+            current_indicator.get_root_window ().get_geometry (null, null, out wingpanel_width, null);
+
+            allocation.x += indicator_allocation.x +
+                            container_allocation.x -
+                            ((allocation.width - indicator_allocation.width) / 2);
+
+            if (allocation.x < 0) {
+                allocation.x = 0;
+            }
+
+            if (allocation.x + allocation.width > wingpanel_width) {
+                allocation.x = wingpanel_width - allocation.width;
+            }
+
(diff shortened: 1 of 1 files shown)


src/Services/PopoverManager.vala
```diff
@@ -102,6 +102,27 @@ public class Wingpanel.Services.PopoverManager : Object {
...
             popover.get_allocation (out allocation);
 
+            Gtk.Allocation indicator_allocation;
+            current_indicator.get_allocation (out indicator_allocation);
+
+            Gtk.Allocation container_allocation;
+            current_indicator.get_parent ().get_allocation (out container_allocation);
+
+            int wingpanel_width;
+            current_indicator.get_root_window ().get_geometry (null, null, out wingpanel_width, null);
```
donadigo: Can't we just use the already available `owner` field here?

davidmhewitt: Thanks for the review, @donadigo 

Since `get_root_window` returns a `Gdk.Window`, not a `Gtk.Window`. We still need the call to `get_root_window`. However, we still get the same resulting window if we get it through the `owner` and it may be slightly more efficient, so I've push a commit that switches to that.

Thanks.

PR #68, opened 2017-11-30
Use both color and translucent classes at once

danrabbit:
This makes it possible to select wingpanel with `.translucent.color-light` so that we can have a white panel on bright backgrounds.

Test with: https://github.com/elementary/stylesheet/pull/236

Please test with a range of wallpapers. The luminescence threshold must be lower for busy backgrounds or it will not be used when appropriate.


cassidyjames:
Should probably tweak the threshold a bit so that wallpapers like this get the light panel: https://unsplash.com/photos/dClHqW-EfS8

Currently with this branch:
image

PR #71, opened 2017-12-08
README.md: Fix translation badge’s link

fitojb:


danrabbit:
I think just doing the project might be better than doing the component in this case since it can be confusing to contributors that wingpanel itself does not contain the indicators

PR #73, opened 2017-12-09
Drag from panel

### LINKED ISSUE #8: Drag from the panel to unmaximize a window
This is a feature requested by people coming from GNOME Shell. They expect that if you start a drag operation from the panel that it will unmaximize the window

Launchpad Details: [#LP1589261](https://bugs.launchpad.net/bugs/1589261) Daniel Fore - 2016-06-05 15:21:21 +0000

donadigo:
Fixes #8.

Makes it possible to drag the maximized window from the panel itself.
This should work for both half-screen maximized windows (only the area above them) and for fully maximized windows.

It would be nice if someone could test this with touch as I don't know if it works or not.

### DIFF
--- a/src/Services/BackgroundManager.vala
+++ b/src/Services/BackgroundManager.vala
@@ -32,6 +32,7 @@ namespace Wingpanel.Services {
         public abstract void initialize (int monitor, int panel_height) throws IOError;
         public abstract void remember_focused_window () throws IOError;
         public abstract void restore_focused_window () throws IOError;
+        public abstract bool begin_grab_focused_window (int x, int y, int button, uint time, uint state) throws IOError;
     }
 
     public class BackgroundManager : Object {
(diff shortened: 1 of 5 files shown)


Philip-Scott:
### The good: 
- Code looks good, as always ;) 
- Works with the mouse

### The bad
- Doesn't work with a Touch Screen (Although touch input in X is fucky, but will investigate if it's something on here: https://github.com/elementary/wingpanel/pull/73/files#diff-4a662e44624e7cf9aed4775f16f636daR71 ) 
- (And would require Gala code) Since you only get the focused window, if there's a maximized window behind the focused one, you are not able to grab the window from the panel

khurshid-alam:
@donadigo Remove gpg signature from the commit first.

donadigo:
@Philip-Scott changes applied. Ready for another review.

Philip-Scott:
And it also crashes if no window on any desktop is maximized

donadigo:
@Philip-Scott fixed Gala crashes (hopefully), and added conditional compiling for the deprecated method.

PR #79, opened 2018-01-04
Switch.vala: rewrite in GObject style

danrabbit:
This breaks API a lot


jeremypw:
Why not maintain the old interface (e.g. getter functions) as alternatives but mark as deprecated until sure no indicators using them?

PR #88, opened 2018-01-29
Fix 3rd party indicator ordering

davidmhewitt:
@stsdc is developing a system monitor wingpanel indicator and it was appearing between the sound and keyboard indicators.

zrzut_ekranu_z_2018-01-29_17 56 59

Because the keyboard indicator was at position 0 of the array, it was being alphabetically sorted with the ayatana indicators and any other indicators that came along and weren't defined.

I've taken this opportunity to add "codenames" for both of those categories so we can ensure a consistent sort order with the system indicators sorted after everything else.

This maintains ayatana compatibility for now in case we want to release a new version of wingpanel in the mean time. I assume another PR will come along and rip out the ayatana support at some point :smile:

### DIFF
--- a/src/Services/IndicatorSorter.vala
+++ b/src/Services/IndicatorSorter.vala
@@ -25,8 +25,8 @@
  * be passed as CompareFuncs.
  */
 public class Wingpanel.Services.IndicatorSorter : Object {
-    private const string UNKNOWN_INDICATOR = "unknown";
-    private const string AYATANA_INDICATOR = "ayatana";
+    private const string UNKNOWN_INDICATOR = "xxx-unknown";
+    private const string AYATANA_INDICATOR = "xxx-ayatana";
 
     /* The order in which the indicators are shown from left to right. */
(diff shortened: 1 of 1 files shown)


src/Services/IndicatorSorter.vala
```diff
@@ -25,18 +25,24 @@
...
-        Indicator.SOUND,
-        Indicator.NETWORK,
-        Indicator.BLUETOOTH,
-        Indicator.PRINTER,
-        Indicator.SYNC,
-        Indicator.POWER,
-        Indicator.MESSAGES,
-        Indicator.SESSION
-    };
+    private static Gee.HashMap<string, int> INDICATOR_ORDER = new Gee.HashMap<string,int> ();
```
donadigo: Don't use `UPPER_CASE` for static members, looks weird when used.

davidmhewitt:
@donadigo Thanks for the review! done

PR #125, opened 2018-06-28
Add the option to run wingpanel as a greeter type

danrabbit:
Adds a commandline arg to run wingpanel as a greeter type server. This would allow us to use wingpanel proper in the greeter instead of carrying a custom panel there

### DIFF
--- a/src/Wingpanel.vala
+++ b/src/Wingpanel.vala
@@ -34,7 +34,7 @@ namespace Wingpanel {
         private const OptionEntry[] OPTIONS = {
             { OPEN_INDICATOR_ACTION_NAME, 'o', 0, OptionArg.STRING, null, "Open an indicator", "code_name" },
             { CLOSE_INDICATOR_ACTION_NAME, 'c', 0, OptionArg.STRING, null, "Close an indicator", "code_name" },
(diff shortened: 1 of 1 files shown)


src/Wingpanel.vala
```diff
@@ -28,11 +28,13 @@ namespace Wingpanel {
         private const string LIST_INDICATORS_ACTION_NAME = "list-indicators";
         private const string OPEN_INDICATOR_ACTION_NAME = "open-indicator";
         private const string CLOSE_INDICATOR_ACTION_NAME = "close-indicator";
+        private const string SERVER_TYPE_ACTION_NAME = "greeter";
         private const string TOGGLE_INDICATOR_ACTION_NAME = "toggle-indicator";
 
         private const OptionEntry[] OPTIONS = {
             { OPEN_INDICATOR_ACTION_NAME, 'o', 0, OptionArg.STRING, null, "Open an indicator", "code_name" },
             { CLOSE_INDICATOR_ACTION_NAME, 'c', 0, OptionArg.STRING, null, "Close an indicator", "code_name" },
+            { SERVER_TYPE_ACTION_NAME, 'g', 0, OptionArg.NONE, null, "server is a greeter", "code_name" },
```
N00byKing: Shouldn't the `s` in `server` be capitalized?

tintou: `"code_name"` should be `null` here too

Issue #92, opened 2018-02-27
Struts are half the width on HiDPI


cassidyjames:
It looks like Wingpanel's struts are only working on the left side of my HiDPI display; at the exact middle, windows no longer bump up against the bottom of the panel, and sometimes skinny windows are placed under the right side of the panel.

cassidyjames:
The strut appears to occupy the red highlighted area here:

![wingpanel-strut-hidpi](https://user-images.githubusercontent.com/611168/36742004-d008adf8-1ba3-11e8-95e4-395a8d035075.png)

You can even bump a skinny window up against the right side to see the rightmost edge of the strut.

jlnr:
Maybe related? Wingpanel reserves more vertical space than it needs on HiDPI, which leads to gaps between fullscreen windows and Wingpanel:

![screenshot von 2018-04-26 09 24 28 2x](https://user-images.githubusercontent.com/70772/39291657-024af862-4934-11e8-994c-64304f059b84.png)

PR #131, opened 2018-07-08
Fix the width of struts on HiDPI displays

JosephMcc:
The scale factor is being accounted for when setting the height but
not the width. Be sure it's set for both.

Fixes: https://github.com/elementary/wingpanel/issues/92


andreasfelix:
I think the `scale_factor` is also needed for the `strut-top-start-x`. Otherwise one also reserves space on a secondary monitor if the primary is the right one.

Issue #134, opened 2018-07-09
Dragging on the panel always selects left window on HiDPI


cassidyjames:
On my HiDPI display, dragging on the panel to until always moves the left-most window. I am guessing we're not multiplying the drag geometry by the scale factor?

Issue #133, opened 2018-07-09
Don't initiate focus/drag on click/release


cassidyjames:
When you have tiled windows and click/release on a blank spot on the panel, currently it initiates a drag and resizes/untiles the app. Instead, we should wait to do that until there's an actual drag.

JosephMcc:
I can't replicate this. For me when I click on an empty spot of the panel over a tiled window it switches to a hand but doesn't do anything until I drag completely down off the panel. That is running in normal DPI.

I have two patches applied: https://github.com/elementary/wingpanel/commit/8559b1daff0948f18daf06c2b94dddda68639642 and https://github.com/elementary/gala/pull/288 but I'm not sure if they would affect that.

danrabbit:
I'm not seeing that the app is untiled but I do see that the cursor is changing and focus is being dealt out with only a tap/click

JosephMcc:
What's the desired behavior here? Should the mouse cursor change? And when? When should focus actually change?

danrabbit:
I think like Cass said it should be only after a drag has begun and not just after a click. I'm not sure how easy that is to differentiate with Gtk+ though.

I'm going to unmilestone this because it's kind of an inconvenience but I don't think we should withhold release over it

danrabbit:
So with latest the behavior I'm seeing is that focus is assigned to the app about to be dragged and the panel no longer takes focus. I'm also not seeing any untiling behavior. Is this fixed?

cassidyjames:
@danrabbit the problem was never that the panel took focus, it's that the panel forces a window to grab focus even if the user hasn't dragged at all. The issue where it untiles apps has been fixed but the focus grabbing is still annoying especially if you have a floating window on top of a maximized window, and barely misclick an indicator: your maximized window steals focus and it's not at all clear why it happened.

PR #135, opened 2018-07-10
Fix dragging from panel on HiDPI displays

JosephMcc:
The coordinates returned from the event need to be multiplied by
the scale factor.

Fixes: https://github.com/elementary/wingpanel/issues/134


JosephMcc:
In testing when using this patch I can no longer reproduce this issue either: https://github.com/elementary/wingpanel/issues/133 at least under HiDPI

PR #139, opened 2018-08-23

torikulhabib:
Fix gereeter Balck panel
https://github.com/elementary/greeter/issues/132


src/PanelWindow.vala
```diff
@@ -132,6 +132,9 @@ public class Wingpanel.PanelWindow : Gtk.Window {
         var monitor = monitor_number == -1 ? this.screen.get_primary_monitor () : monitor_number;
         var position_top = monitor_y - panel_displacement;
         var scale_factor = this.get_scale_factor ();
+        if ("io.elementary.greeter" != "fullscreen"){
```
Djaler: What did you want to do with this line?

torikulhabib: I tried to make an exception to the greeter so that the wingpanel doesn't press greeter down.

PR #140, opened 2018-08-25

torikulhabib:
Please made code to be simple @donadigo


torikulhabib:
this depend with pull https://github.com/elementary/wingpanel/pull/139

PR #141, opened 2018-09-08

torikulhabib:
close https://github.com/elementary/wingpanel/pull/139
close https://github.com/elementary/wingpanel/pull/140


torikulhabib:
close because simple https://github.com/elementary/greeter/pull/154

PR #180, opened 2018-12-06
RDNN exec and file names

### LINKED ISSUE #4: use the right reversed elementary Scheme
We do not own pantheon.org, so our GSettings living under org.pantheon does not follow RDNN (https://en.wikipedia.org/wiki/Reverse_domain_name_notation).

In Loki+1, we should switch to using io.elementary for shell/OS components, and com.github.elementary for first-party apps.

Things to update:
 * Application id
 * Desktop filename
 * AppStream filename and id
 * GSettings path and id
 * DBus interfaces

Launchpad Details: [#LP1594598](https://bugs.launchpad.net/bugs/1594598) Cassidy James Blaede - 2016-06-20 22:26:20 +0000

danrabbit:
Fixes #4


danrabbit:
It looks like we previously decided to use `io.elementary.desktop.wingpanel` for the gsettings location so we should probably use that

tintou:
Note that this is changing the gettext domain to `io.elementary.panel` so the .pot file should be renamed in git too and Weblate should be fixed accordingly.

Issue #90, opened 2018-02-10
Throw an "Agent" before adding a new indicator


danrabbit:
We should throw an agent-style dialog if an app tries to add an indicator and make sure we only add indicators with a user's consent

Djaler:
I hope it will be one-per-app dialog?

cassidyjames:
@Djaler yeah I think that makes sense.

cassidyjames:
ASCII sketch time:

```
|--------------------------------------------------|
|                                                  |
|  ICON  Allow an Indicator from App Name?         |
|  ICON                                            |
|        This will add an icon to the panel, even  |
|        when no App Name window is open. This can |
         be changed from the Desktop settings.     |
|                                                  |
|                                [ Deny ] [ Alow ] |
|--------------------------------------------------|
```

Thoughts on the exact language?

danrabbit:
Closing since indicators aren't a thing anymore

PR #186, opened 2019-01-10
Indicator: deprecate unused properties

danrabbit:
These have never been used, so maybe time to deprecate?


cassidyjames:
I think we'll need these if we want to do #90, right?

danrabbit:
Naw I think the path forward is to use the metadata from AppStream so we have one canonical source of data

PR #187, opened 2019-01-11
Fix revealer using unmap event

peteruithoven:
Fixes: elementary/wingpanel-indicator-bluetooth#64
Supersedes: #174

After the bluetooth indicator disappears it sometimes doesn't properly re-appear. 
```
systemctl restart bluetooth
```

I can also reproduce the issue by rapidly switching nightlight on off. You see it sometimes not showing up or even showing only half of the icon:
screen record from 2019-01-11 21 20 15
screen record from 2019-01-11 21 19 21
(Using <kbd>Enter</kbd> to switch the button makes this easier)

I noticed I could fix this by commenting the delayed re-ordering:
https://github.com/elementary/wingpanel/blob/master/src/Widgets/IndicatorEntry.vala#L82-L85
But this means the indicator isn't removed and you'll see extra margins. 

Removing the previous timeout and increasing the interval helps a great deal. But that means another arbitrary value, which could still give issues on slower systems and adds a delayed sudden jump in the indicators. 

I was hoping to find a solution like: https://github.com/elementary/wingpanel/pull/155

This PR's uses the `unmap` signal to know a indicator is hidden, it then start a 0ms timeout to request the menubar reordering. The timeout makes it cancelable when the indicator is quickly made visible again.

Still feels like quite a workaround, I'm open for better solutions. 
Disconnecting from the `unmap` signal when the indicator is made visible again gave issues for some reason. 
It could be that a better solution is doing a less crude reordering in the menubar?

### DIFF
--- a/src/Widgets/IndicatorEntry.vala
+++ b/src/Widgets/IndicatorEntry.vala
@@ -38,8 +38,6 @@ public class Wingpanel.Widgets.IndicatorEntry : Gtk.MenuItem {
 
     Services.PopoverManager popover_manager;
 
-    private uint reorder_timeout_id;
-
     public IndicatorEntry (Indicator base_indicator, Services.PopoverManager popover_manager) {
         this.popover_manager = popover_manager;
         this.base_indicator = base_indicator;
@@ -78,10 +76,6 @@ public class Wingpanel.Widgets.IndicatorEntry : Gtk.MenuItem {
                     popover_manager.register_indicator (this);
                     menu_bar.apply_new_order ();
                     set_reveal (base_indicator.visible);
-                    // cancel unmapped reorder
-                    if (reorder_timeout_id > 0) {
-                        Source.remove (reorder_timeout_id);
-                    }
                 } else {
                     set_reveal (base_indicator.visible);
                     popover_manager.unregister_indicator (this);
@@ -128,12 +122,7 @@ public class Wingpanel.Widgets.IndicatorEntry : Gtk.MenuItem {
 
     private void indicator_unmapped () {
         base_indicator.get_display_widget ().unmap.disconnect (indicator_unmapped);
-
-        // using timeout to make it cancelable
-        reorder_timeout_id = Timeout.add (0, () => {
-            menu_bar.apply_new_order ();
-            return false;
-        });
+        menu_bar.apply_new_order ();
     }
 
     public void set_transition_type (Gtk.RevealerTransitionType transition_type) {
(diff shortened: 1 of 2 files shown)


jeremypw:
I'll need a little time to get up to speed with the code base before reviewing ...

jeremypw:
I can confirm that it fixes the problem - I'll have a think about whether a more elegant solution exists.

peteruithoven:
I'm afraid that when I remove the timeout stuff and change the `apply_new_order` to the following, I still have the issue that the indicator sometimes doesn't appear of even only appears half way. 
```
    public void apply_new_order () {
        Idle.add(() => {
            clear ();
            append_all_items ();
            return false;
        });
    }
```
Maybe there is a better way to make the reorder call cancelable than with a timeout? 
Like I mentioned I tried disconnecting from the unmap when the indicator was made visible again, but that caused the Wingpanel to crash when enabling, disabling quickly.

jeremypw:
@peteruithoven:  OK,  I guess it must depend on hardware characteristics - I could not trigger the bug using a simple Idle (but I could only occasionally trigger it in master anyway).  You could try using Idle.add_full with GLib.Priority.LOW and/or also keep a reference to the source id so that you can cancel the Idle if apply_new_order is called too rapidly.  It seems that the fix should be kept within IndicatorMenubar if possible.

jeremypw:
Don't forget to install the changes and restart wingpanel - I have sometimes forgotten!

peteruithoven:
> You could try using Idle.add_full with GLib.Priority.LOW and/or also keep a reference to the source id so that you can cancel the Idle if apply_new_order is called too rapidly.

I'm happy to report that that seems to work. Did I implement it like you meant it? 

I kind of prevent that mistake by temporarily removing wingpanel from the monitored list of cerbere and using the following the following command to compile and run:
`ninja && sudo ninja install &&  G_MESSAGES_DEBUG=all  wingpanel`

jeremypw:
@peteruithoven :  Could you try this?

```
        base_indicator.notify["visible"].connect (() => {
            if (menu_bar != null && (revealer.get_reveal_child () != base_indicator.visible)) {
                /* order will be changed so close all open popovers */
                popover_manager.close ();

                if (base_indicator.visible) {
                    popover_manager.register_indicator (this);
                } else {
                    popover_manager.unregister_indicator (this);
                }

                set_reveal (base_indicator.visible);
                menu_bar.apply_new_order ();
            } else {
                set_reveal (base_indicator.visible);
            }
        });
```
It works for me.

jeremypw:
Ah! I see you have already pushed some changes before I edited my comment - sorry.

peteruithoven:
@jeremypw thanks I forgot about that error, I've added a check.

You're last suggestion prevents the Revealer hide animation, because it's removed immediately?

jeremypw:
Ah, OK.  I'll approve as is then.

Issue #130, opened 2018-07-07
Window opens/can slide below wingpanel


arshubham:
Steps:
1. Start a session.
2. Log out.
3. Log back in.

Affects all applications and reproducible every time.

Screenshots:
- Un-maximized window
![screenshot from 2018-07-07 13-42-31](https://user-images.githubusercontent.com/14235191/42408759-7a4e39bc-81ee-11e8-93a5-24d1da6528ef.png)

- Maximized window
![screenshot from 2018-07-07 13-42-46](https://user-images.githubusercontent.com/14235191/42408761-82c09aea-81ee-11e8-86ba-f5eb53a0e0ce.png)
Note that the titlebar is shorter in height.

janxkoci:
I have the same issue, but only sometimes, and I haven't found a trigger yet. After some reboots it's fine, but other times it's like this. I was thinking it appeared after I enter the desktop settings (wallpaper, dock, etc.) but the issue also appears after cold start, at least sometimes.

albergh:
I have a quick fix when that happens, unmaximize all apps, I open Gnome system monitor, search wingpanel, click on End process (don't worry, it will restart itself) and maximize again. That should do the trick untill the next reboot

peteruithoven:
I've been noticing this as well sometimes. 

I'm pretty sure it has something to do with this animation:
https://github.com/elementary/wingpanel/blob/master/src/PanelWindow.vala#L76-L94
If I increase that timeout interval to `1000 / panel_height` it doesn't seem to happen anymore.
```
Timeout.add (1000 / panel_height, animation_step);
```

sprite-1:
@peteruithoven will you add it to the official source code?

peteruithoven:
No sorry, because it increases the length of the animation and it doesn't fix the underlying issue. 
I believe the issues is that it animates and while it's animating it reads the height and uses that to declare it's height to the window manager, sometimes though the animation is delayed giving wrong height readouts.

janxkoci:
I'm a bit surprised that height of wingpanel isn't hardcoded. I'm generally against hardcoding values, but since wingpanel is not meant to be editable, is there any harm?

peteruithoven:
@janxkoci I don't get where you pull that assumption from, it kinda is hardcoded: 
https://github.com/elementary/wingpanel/blob/master/src/Widgets/Panel.vala#L33

[...]

PR #198, opened 2019-02-25
Fix wingpanel potentially overlapping windows

peteruithoven:
Fixes: #130

I noticed that sometimes my wingpanel would overlap left/right tiled windows. Looking into it it seemed like the allocated height of the panel changed from 24 (the requested height) to 30. For some people, in some cases this switch would happen after the [slide in animation](https://github.com/elementary/wingpanel/blob/master/src/PanelWindow.vala#L76-L86) where the [truts](https://github.com/elementary/wingpanel/blob/master/src/PanelWindow.vala#L126-L153)? are updated. 

I filtered out all indicators and it seemed like these indicators caused the panel's height to go from 24 to 30:
- [session](https://github.com/elementary/wingpanel-indicator-session)
- [network](https://github.com/elementary/wingpanel-indicator-network)
- [sound](https://github.com/elementary/wingpanel-indicator-sound)
- [power](https://github.com/elementary/wingpanel-indicator-power/)

You can also see that those 4 have a margin above and below:
screenshot from 2019-02-25 22-49-18

I'm assuming we actually like the 30, so this PR makes sure we start of at 30. 

For comparison:
24 height:
screenshot from 2019-02-25 22-03-15
30 height:
screenshot from 2019-02-25 22-05-10

### DIFF
--- a/src/Widgets/Panel.vala
+++ b/src/Widgets/Panel.vala
@@ -30,7 +30,7 @@ public class Wingpanel.Widgets.Panel : Gtk.EventBox {
     public Panel (Services.PopoverManager popover_manager) {
         Object (popover_manager : popover_manager);
 
-        this.set_size_request (-1, 24);
+        this.set_size_request (-1, 30);
 
         this.hexpand = true;
         this.vexpand = true;


worldofpeace:
@codygarver Can we get a `2.2.3` release with this fix?

Issue #205, opened 2019-04-24
Focus newly started app


peteruithoven:
Digging into [Apps are not in focus after launching from application menu #19](https://github.com/elementary/applications-menu/issues/19) I found this logic:

https://github.com/elementary/wingpanel/blob/25af8fecfd3738f2b7b0f327865e5f72a6a937d5/src/PanelWindow.vala#L155-L168

I'm afraid this steals the focus from the newly launched app. 
Does anyone have any idea how to disable this restore_window request when a new app is launched?

davidmhewitt:
The `remember_window`, `restore_window` logic seems a little weird to me. Surely it would be better to listen for focus changes and always have the "remembered" window be the last focused window excluding wingpanel.

davidmhewitt:
That said, the issue at hand is likely a race condition.

Smaller, faster apps manage to open their window and grab focus before this code restores the (wrong) previously focused window. Whereas slower apps take focus after the previously focused window has been given focus back.

peteruithoven:
Interesting! 

Tricky, as far as I can find on valadoc there is no global focus change listener?
We could listen on the window that currently has focus for [focus_out_event](https://valadoc.org/gtk+-3.0/Gtk.Widget.focus_out_event.html) but I'm not sure that event includes the window that gained the focus.

davidmhewitt:
The `wingpanel-interface` code that handles this remembering of window focus is a Gala plugin. I imagine that a plugin inside the Window Manager might be able to register for window focus gained/lost events, although I'm not 100% sure since it's a while since I looked at Gala.

peteruithoven:
Looking through the mutter bindings there doesn't seem to be a focus changed or even a `focus_out event`. 
https://github.com/elementary/wingpanel/blob/master/vapi/libmutter.vapi
But then again I might not understand how to properly search for this.

PR #206, opened 2019-04-24
Keep better record of focused apps

peteruithoven:
Fixes: #205 and https://github.com/elementary/applications-menu/issues/19

Following @davidmhewitt advice to listen for window focus changes fixing the following hypothesis:

> Smaller, faster apps manage to open their window and grab focus before this code restores the (wrong) previously focused window. Whereas slower apps take focus after the previously focused window has been given focus back.

### DIFF
--- a/wingpanel-interface/FocusManager.vala
+++ b/wingpanel-interface/FocusManager.vala
@@ -33,14 +33,15 @@ public class WingpanelInterface.FocusManager : Object {
 
     public void remember_focused_window () {
         var windows = current_workspace.list_windows ();
-
         foreach (Meta.Window window in windows) {
+            window.focused.connect (window_focused);
+            window.unmanaged.connect (window_unmanaged);
             if (window.has_focus ()) {
                 last_focused_window = window;
-
-                return;
             }
         }
+
+        Main.screen.get_display ().window_created.connect (window_created);
     }
 
     public void restore_focused_window () {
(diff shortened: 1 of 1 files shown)


wingpanel-interface/FocusManager.vala
```diff
@@ -33,21 +33,46 @@ public class WingpanelInterface.FocusManager : Object {
 
     public void remember_focused_window () {
         var windows = current_workspace.list_windows ();
-
         foreach (Meta.Window window in windows) {
+            window.focused.connect (window_focused);
+            window.unmanaged.connect (window_unmanaged);
```
donadigo: Just call `window_created` here.

vjr:
I wonder if this also fixes the intermittent password field focus issue in the current released greeter: https://github.com/elementary/greeter/issues/144

peteruithoven:
I don't think it will. What happened was that:
1. when you open the applications menu the wingpanel where the application menu is apart of) would get focus, but before that would happen it would store the previously focused window. 
2. An app would start, retrieving focus
3. The wingpanel would restore focus to the stored app, not the newly started app.

In some cases, with slower applications 2 and 3 would be reverted and the new app would retrieve focus. 
With this PR we keep recording focus changes so we always focus the last application that had focus. 

I don't think any of this is relevant for the Greeter, new or old.

PR #208, opened 2019-05-08
Fix panel transparency calculations

### LINKED ISSUE #207: Background isn't totally transparent anymore.
When there is no window maximized, wingpanel background is not 100% transparent.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/73404698-background-isn-t-totally-transparent-anymore?utm_campaign=plugin&utm_content=tracker%2F58844522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F58844522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #207 

The new mutter version seems to give us a texture that's 3px in both dimensions bigger than the background actor that the texture is rendered from. This tends to leave a darker border around the texture, which doesn't help the calculations.

Adjust the logic to ignore 2px from every side of the texture we pull from clutter and also fix the following logic errors while we're here:
 - Luminance calculations were being done with reversed RGB values
 - Copy and paste error in argument range check
 - Calculations didn't account for an offset start position other than 0 and would only calculate on an area of (requested size - offset) rather than the requested area

This will need testing with a variety of backgrounds again to make sure it's doing sensible things, but it seems much closer to the old behaviour again.

Remember to install and reboot/restart gala to test as this is a gala plugin.

### DIFF
--- a/wingpanel-interface/Utils.vala
+++ b/wingpanel-interface/Utils.vala
@@ -61,7 +61,11 @@ namespace WingpanelInterface.Utils {
         var bg_actor_width = (int)background.width;
         var bg_actor_height = (int)background.height;
 
-        // Avoid looking at the edges of the texture as it often has a black border
+        // A commit in mutter added some padding to offscreen textures, so we
+        // need to avoid looking at the edges of the texture as it often has a
+        // black border. The commit specifies that up to 1.75px around each side
+        // could now be padding, so cut off 2px from left and top if necessary
+        // (https://gitlab.gnome.org/GNOME/mutter/commit/8655bc5d8de6a969e0ca83eff8e450f62d28fbee)
         int x_start = reference_x;
         if (x_start < 2) {
             x_start = 2;
@@ -72,8 +76,10 @@ namespace WingpanelInterface.Utils {
             y_start = 2;
         }
 
-        // If the caller has specified an area of interest smaller than 2px less than the width/height,
-        // use that, otherwise chop 2px off the edges again
+        // For the same reason as above, we need to not use the bottom and right
+        // 2px of the texture. However, if the caller has specified an area of
+        // interest that already misses these parts, use that instead, otherwise
+        // chop 2px
         int width = int.min (bg_actor_width - 2 - reference_x, reference_width);
         int height = int.min (bg_actor_height - 2 - reference_y, reference_height);
 
--- a/wingpanel-interface/Utils.vala
+++ b/wingpanel-interface/Utils.vala
@@ -58,15 +58,26 @@ namespace WingpanelInterface.Utils {
         var effect = new DummyOffscreenEffect ();
         background.add_effect (effect);
 
-        var tex_width = (int)background.width;
-        var tex_height = (int)background.height;
+        var bg_actor_width = (int)background.width;
+        var bg_actor_height = (int)background.height;
 
+        // Avoid looking at the edges of the texture as it often has a black border
         int x_start = reference_x;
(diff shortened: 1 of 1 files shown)


peteruithoven:
Great work!
While this is a huge improvement, I've compared this branch with an old  iso from 2018-01-16 and there are 2 wallpapers where it gives a different result:
Screenshot from 2019-05-08 20-56-49@2x
Screenshot from 2019-05-08 20-56-33@2x

davidmhewitt:
I personally feel it's giving a better result, the bottom one feels like it needed a background as the contrast was poor on the date/time and I'm indifferent about the top one.

I guess it's for @elementary/ux to decide

peteruithoven:
Trying the `test-background-capture` (changed to use the full `height`) the captures do make sense. So there doesn't seem to be a HiDPI issue. 
test
test (copy)

I actually agree it's an improvement.

kgrubb:
Sorry for the image spam! I took screenshots of one of the default wallpapers as well as a bunch of images under the "wallpapers" section on unsplash to check the behavior (from this branch). I'm using a 16:10 monitor (1440x900) with loDPI settings in the display plug. In my opinion, I also agree that this is an improvement from both pre-mutter-upgrade and post-mutter-upgrade.

Screen Shot 2019-05-08 at 4 23 03 PM
Screen Shot 2019-05-08 at 4 34 40 PM
Screen Shot 2019-05-08 at 4 34 56 PM
Screen Shot 2019-05-08 at 4 35 08 PM
Screen Shot 2019-05-08 at 4 35 19 PM

This next one one is a bit weird due to the tower right in the middle, but I still don't believe its bad enough to change any calculations (looks fine to me for the most part).

Screen Shot 2019-05-08 at 4 35 42 PM
Screen Shot 2019-05-08 at 4 36 02 PM
Screen Shot 2019-05-08 at 4 36 16 PM
Screen Shot 2019-05-08 at 4 36 30 PM
Screen Shot 2019-05-08 at 4 36 39 PM

danrabbit:
I'm happy with the results of this branch. From the discussion in Slack, it sounds like @donadigo might be interested in doing the actual code review :)

micahilbery:
Does this make the default system wallpaper have a top bar?

davidmhewitt:
@micahilbery No, I get a fully transparent wingpanel with this branch and the default wallpaper.

`master` seemingly has a semi-transparent black background at the moment, which looking at the screenshots above is consistent with what it was pre-mutter updates too.

kgrubb:
@micahilbery here's a screenshot of the default wallpaper via this branch:

Screen Shot 2019-05-09 at 12 54 22 PM


Something that I noticed (definitely not an issue, its not wingpanel related, but greeter related), the greeter does not follow the same color calculations.

Here's an example of a wallpaper using this branch's wingpanel calculations (honestly, beautiful!):

Screen Shot 2019-05-09 at 12 49 31 PM

Here's greeter's panel (it shows up as white here):

Screen Shot 2019-05-09 at 12 49 51 PM

Here's why i think its entirely a non-issue though. With the new greeter (master), the white coloring makes more sense to always be white:

Screen Shot 2019-05-09 at 12 52 51 PM

yusufsetyaa:
kinda obsolete for now, since I'm new to eleOS, where is the exact location of this file so I can edit it?

PR #93, opened 2018-03-04
Autohide: optionally hide when the mouse is not over panel or indicator menus

quequotion:
I think it is worth [reconsidering](https://bugs.launchpad.net/wingpanel/+bug/1079575) autohide for [wingpanel](https://www.bountysource.com/issues/955498-add-optional-hide-mode) as an optional feature.

In fact many applications do not "fullscreen" when told to "maximize" and the presence of the panel enforces a 24px distinction between the two. There are also applications that attempt to draw actionable areas under the panel. Users could enable autohide to have their full resolution available and work around apps that do not play well with panels.

This patch adds an autohide mode and gsetting (org.pantheon.desktop.wingpanel autohide) with the behavior OFF as default.

PR #212, opened 2019-05-14
Add autohide support for wingpanel

### LINKED ISSUE #53: Auto-hide wingpanel
Maybe it would be possible to integrate a auto-hide function for the next version of elementary OS ?

Thus the applications could be displayed in full screen mode.

Best regards.

donadigo:
Depends on https://github.com/elementary/gala/pull/535.
Fixes #53.
This PR adds an initial support for autohide in wingpanel. 

There have been multiple attempts to add an autohide and some of them really made this feature work, like #93. This one in particular made me look at helping / implementing this. While #93 introduces many new autohide modes, it wouldn't be quite ready to merge, there are issues with animations and it introduces libraries we wouldn't want to use in the future. I originally based this branch on this PR but quickly realized that I need to rewrite most of the code to make it something that can be mergable to master.

This one adds two autohide modes: float and dodge float which work similarly to #93. Float always hides the panel and shows it when the mouse enters the top of the screen and dodge float hides the panel only when the current window is maximized.

It doesn't add any new libraries and it uses the already existing wingpanel interface for the new stuff that's needed. I tried to keep the code understandable and made comments where appropriate.

Current issues:
- [x] Gala needs to know about the panel extending and it needs to offset UI elements like the notification stack
- [x] Needs tweaking the delay value for a good default. I think something like ~250 would be sufficient.


tintou:
I, personally, do not like hidden features

donadigo:
@tintou I do not intend this to be a hidden feature. If this was it, then I wouldn't bother. I think that the elementary team at some point said that they could adopt this feature and expose it to the UI, just like plank's settings, if it was worked well enough.

quequotion:
Exicitig news!

Great to see this come full circle--it was an IRC conversation with you that got me started on #93!

I'll try to make some PKGBUILDs to test this locally; would be good to have a side-by-side comparison of their behavior. 

I'll try porting any improvements from yours to mine that I can, but it may end up that I keep a version of my fork, using libwnck, around for non-Gala Pantheon derivatives.

n-osennij:
any updates?

miere:
Hi guys, any feedback on this?
Need any help with the implementation?

danirabbit:
Gonna close this since it now conflicts a lot and hasn't had any movement in a few years. Additional thoughts on the issue report

PR #227, opened 2019-07-11
Disable GtkWidget-window-dragging on panel

### LINKED ISSUE #169: Wingpanel moves
The wingpanel bar moves from the top center for some reason. Haven't found what triggers it, but it has happened every day since I installed Juno. It always moves to the exact same spot and stays there unless I reboot.

I've tried to move it with my mouse and touchscreen but it stays there.  

I've attached a screenshot.

Thanks for your help.

screenshot from 2018-10-25 17 00 30

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/65287019-wingpanel-moves?utm_campaign=plugin&utm_content=tracker%2F58844522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F58844522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

davidmhewitt:
Fixes #169 

We were inheriting `-GtkWidget-window-dragging: true;` from `Gtk.STYLE_CLASS_MENUBAR` which was causing the issue of being able to drag the panel around with 2+ fingers on a touchscreen.

I don't know if we want to do this in a neater way (i.e. in a GResource or by doing something in the stylesheet), but this seems to be the cause of the issue and fixes it for me.

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -7,3 +7,9 @@ install_data(
     'wingpanel.svg',
     install_dir: join_paths(get_option('datadir'), 'icons', 'hicolor', 'scalable', 'apps')
 )
+
+# Use GResource
+css_gresource = gnome.compile_resources(
+    'gresource_css',
+    meson.project_name() + '.gresource.xml'
+)
--- a/data/styles/application.css
+++ b/data/styles/application.css
@@ -1,0 +1,3 @@
+.panel {
+    -GtkWidget-window-dragging: false;
+}
--- a/data/wingpanel.gresource.xml
+++ b/data/wingpanel.gresource.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<gresources>
+  <gresource prefix="/io/elementary/wingpanel">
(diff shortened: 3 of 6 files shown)


danrabbit:
Nice job!

I've been thinking about moving the essential panel styles here for a while. If you wanna throw this in a gresource, I can do a follow up commit with other stuff too

peteruithoven:
I can also confirm this solves it. :tada

CSstefan:
I saw the fix pushed to master, but when does that push out to updates? I've updated terminals and it still moves.

davidmhewitt:
New versions are released manually once a package is considered stable in the daily repositories. So there is no release schedule as such.

However if you wish to test the version of wingpanel incorporating this fix, you can install this Deb: wingpanel_2.2.5+r272+pkg42~daily~ubuntu5.0.1_amd64.deb

You can then restore the original version with `sudo apt install --reinstall wingpanel`

Issue #228, opened 2019-07-15
Shut Down and Log Out dialogs lose focus


cassidyjames:
Filed here and not on the session indicator because this is probably due to the self-unfocusing code of WingPanel?

peteruithoven:
indeed I'm afraid so. 

But, maybe instead of coming up with a workaround for this use case, we can skip the dialog and add the individual buttons directly in the Popover? Just like this mockup from @lainsce: 
![session-indicator-popover_redesign](https://user-images.githubusercontent.com/523210/61255845-a64dcf80-a76a-11e9-9b6f-555ec9ca0267.png)

Blast-City:
In the new dialog it should be possible to cancel with the Esc key.

Right now this is not possible.

cassidyjames:
@Blast-City that's an entirely separate issue, I've filed it at https://github.com/elementary/wingpanel-indicator-session/issues/82

PR #229, opened 2019-07-16
Fix focus on dialogs started from wingpanel

peteruithoven:
Fixes: #228

### DIFF
--- a/wingpanel-interface/FocusManager.vala
+++ b/wingpanel-interface/FocusManager.vala
@@ -64,11 +64,8 @@ public class WingpanelInterface.FocusManager : Object {
     }
 
     void window_focused (Meta.Window window) {
-        switch (window.window_type) {
-            case Meta.WindowType.NORMAL:
-            case Meta.WindowType.DIALOG:
-                last_focused_window = window;
-                break;
+        if (window.window_type != Meta.WindowType.DOCK) {
+            last_focused_window = window;
         }
     }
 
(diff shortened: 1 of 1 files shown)


peteruithoven:
@cassidyjames did you also restart Gala after compiling an restarting wingpanel? (This logic depends on a Gala plugin)

cassidyjames:
@peteruithoven hm, I did not since this looked self-contained within the panel. I'll reboot and continue testing :+1:

peteruithoven:
If you don't want to restart you can also execute `gala -r` from the Applications Menu.

cassidyjames:
@peteruithoven I know, but I figured I'd be safe and restart _everything._ :wink: Now it appears to properly retain focus, but _too_ well: closing the dialog keeps the focus on WingPanel instead of returning it to the previous app. :sweat_smile:

peteruithoven:
Wingpanel && focus is such a nightmare... It stores the `last_focused_window` to restore focus to later on, in this case that became the dialog, which means there is no more record of the previously stored focused window. We'd need to add some kind of exception in probably the FocusManager. Figure out where we want to trigger this manual refocus from when the dialog is closed.

peteruithoven:
I've added an exceptions for windows that are dialogs, making sure we don't lose the last focused window. When the wingpanel requests to restore focus it focuses this dialog and when the dialog is closed it focuses the last focused (non dialog) window.

jeremypw:
Works as expected on dialogs - but not popovers (but I guess that is a different issue).

PR #257, opened 2019-10-16
WingpanelInterface: Set durations with constants

### LINKED ISSUE #250: Replace Granite.Services.Settings with GLib.Settings

danrabbit:
Fixes #250


jeremypw:
I notice that all three gala settings accessed here - `snap-duration`, `minimize-duration` and `workplace-switch-duration` - are all marked as "deprecated and ignored" .  So we need to keep them here?

danrabbit:
You're right! This was done in https://github.com/elementary/gala/pull/513

Updated to use constants.

wingpanel-interface/BackgroundManager.vala
```diff
@@ -105,31 +108,31 @@ public class WingpanelInterface.BackgroundManager : Object {
...
         });
 
         window.notify["minimized"].connect (() => {
-            check_for_state_change (AnimationSettings.get_default ().minimize_duration);
+            check_for_state_change (MINIMIZE_DURATION);
         });
 
         window.workspace_changed.connect (() => {
-            check_for_state_change (AnimationSettings.get_default ().minimize_duration);
+            check_for_state_change (MINIMIZE_DURATION);
```
jeremypw: This is correctly based on master, but is it right that the `workspace_changed` signal closure uses `MINIMIZE_DURATION" and not`"WORKPLACE_SWITCH_DURATION`?   This can be left for another PR (if required).

danrabbit: Oh wow. Thank you for actually reading this! I feel dumb for not noticing this :sweat_smile:

PR #260, opened 2019-10-30
add support for mutter 3.3x

decathorpe:
This is mostly stuff adapted from the corresponding [gala/wip/mutter-3.3x branch](https://github.com/elementary/gala/tree/wip/mutter-3.3x). I had to look up some API changes which didn't affect gala. Anyway, there's no guarantee that this actually works correctly, but it kinda does on fedora 31.

I confirmed that this continues to compile with mutter 3.28, and it does compile on fedora 31 (mutter 3.34.1) and fedora rawhide (mutter 3.35.1). So the support for mutter 3.30 and 3.32 is untested, but present and should be working as well.

**Note 1**: I'm using conditional compilation as it's used elsewhere for this kind of "dual" / "multiple" support, so the original code (mutter 3.28 support) is effectively unchanged.

**Note 2**: This makes wingpanel actually launch in a Pantheon session with gala compiled against mutter 3.34. But gala is still crashing at launch for various reasons I haven't been able to completely understand yet. Either it crashes in workspace handling code, or in screensaver registering code, depending on whether Pantheon is launched from LightDM or GDM, respectively.

Anyway. Treat this as a proof-of-concept starting point.

### DIFF
--- a/wingpanel-interface/FocusManager.vala
+++ b/wingpanel-interface/FocusManager.vala
@@ -25,9 +25,16 @@ public class WingpanelInterface.FocusManager : Object {
     private Meta.Window? last_focused_dialog_window = null;
 
     public FocusManager () {
+#if HAS_MUTTER330
+        unowned Meta.WorkspaceManager manager = Main.display.get_workspace_manager ();
+        manager.workspace_switched.connect (() => {
+            update_current_workspace ();
+        });
+#else
         Main.screen.workspace_switched.connect (() => {
             update_current_workspace ();
         });
+#endif
 
         update_current_workspace ();
     }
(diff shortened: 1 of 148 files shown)


wingpanel-interface/FocusManager.vala
```diff
@@ -87,10 +106,18 @@ public class WingpanelInterface.FocusManager : Object {
...
     public bool begin_grab_focused_window (int x, int y, int button, uint time, uint state) {
+#if HAS_MUTTER330
+        var display = Main.display;
+#else
         var display = Main.screen.get_display ();
+#endif
         var window = display.get_focus_window ();
         if (window == null || !get_can_grab_window (window, x, y)) {
+#if HAS_MUTTER330
+            unowned Meta.Workspace workspace = Main.display.get_workspace_manager ().get_active_workspace ();
```
tintou:
```suggestion
            unowned Meta.Workspace workspace = display.get_workspace_manager ().get_active_workspace ();
```

PR #274, opened 2020-01-10
Add autostart.desktop

danrabbit:
Add an autostart.desktop like we do in Onboarding

### DIFF
--- a/data/autostart.desktop
+++ b/data/autostart.desktop
@@ -1,0 +1,14 @@
+[Desktop Entry]
+Name=Wingpanel
+Comment=Top panel that holds indicators and the applications menu
+Exec=wingpanel
+Icon=io.elementary.wingpanel
+Terminal=false
+Type=Application
+Categories=Utility;
+NoDisplay=true
+X-GNOME-Autostart-Notify=false
+X-GNOME-AutoRestart=true
+X-GNOME-Autostart-enabled=true
+X-GNOME-Autostart-Phase=Panel
+OnlyShowIn=Pantheon;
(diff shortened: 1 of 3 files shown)


danrabbit:
Hm so after testing for a few days I'm noticing that gnome session complains about the panel failing to register itself and it seems like other autostarts (like polkit) are failing.

So there might be more work required here

davidmhewitt:
@danrabbit When you say the polkit autostart is failing, you mean failing to autostart altogether or just failing to register like the panel?

danrabbit:
@davidmhewitt failing to start altogether. My guess is that the panel failing to register stops GNOME Session from performing any "app" class autostarts

davidmhewitt:
@danrabbit Can you confirm that it's not cerbere causing this? As the original indication that we had something wrong with cerbere was the polkit agent not starting...

Stick `.disabled` on the end of your cerbere autostart filename and see what happens. You'll have to start plank manually in your session, or make an autostart file for that too.

danrabbit:
@davidmhewitt Okay so I disabled cerbere and plank is not spawning (as intended).

I can confirm that neither polkit nor appcenter nor any other app-level autostarts are running.

What's interesting though is that after 90 seconds they do load. So I think there must be some kind of timeout here

davidmhewitt:
Yes, 90 seconds is how long `gnome-session` waits for each phase of the autostart, so this makes sense.

It's odd that I didn't seem to notice that issue on my system though. Looks like we need to stick session management registration code in wingpanel and plank after all :slightly_smiling_face: 

Although, given the problems that the polkit dialog causes when it doesn't launch, it should probably be in an earlier phase too.

danrabbit:
This looks good! Whether cerbere launches wingpanel or not I'm not running into issues with Polkit. Seems like this works as expected :)

PR #295, opened 2020-03-06
Fix windows below wingpanel for some 3 monitor layouts

andreasfelix:
Fixes https://github.com/elementary/gala/issues/737


src/PanelWindow.vala
```diff
@@ -212,7 +212,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
             var end_x = rect.x + rect.width - 1;
             if (monitor_x > end_x) {
                 var end_y = rect.y + rect.height - 1;
-                if (panel_end >= rect.y && panel_start <= end_y) {
+                if (panel_start <= end_y && panel_end >= rect.y) {
```
andreasfelix: This lines does not change any logic. I just swap the order for consistency.

src/PanelWindow.vala
```diff
@@ -189,7 +189,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
         var monitor_end_x = monitor_x + monitor_width - 1;
         foreach (unowned Gdk.Rectangle? rect in other_rects) {
             var end_y = rect.y + rect.height - 1;
-            if (end_y > monitor_y) {
+            if (monitor_y > end_y) {
```
andreasfelix: This line fixes the reported bug

src/PanelWindow.vala
```diff
@@ -232,9 +232,9 @@ public class Wingpanel.PanelWindow : Gtk.Window {
         var panel_end = monitor_y + panel_height - 1;
 
         foreach (unowned Gdk.Rectangle? rect in other_rects) {
-            if (monitor_end_x < (rect.x + rect.width - 1)) {
+            if (monitor_end_x < rect.x) {
```
andreasfelix: This fixes another bug

cassidyjames:
And just to confirm, this does not regress the struts issue on multiple-monitor setups? :sweat_smile: I don't have that many displays to properly test this.

src/PanelWindow.vala
```diff
@@ -232,9 +232,9 @@ public class Wingpanel.PanelWindow : Gtk.Window {
...
                 var end_y = rect.y + rect.height - 1;
-                if (panel_end >= rect.y && panel_start <= end_y) {
+                if (panel_start <= end_y && panel_end >= rect.y) {
```
andreasfelix: This line does not change any logic. I just swap the order for consistency.

andreasfelix:
@cassidyjames This just fixes a logical error in the `has_no_monitor_above` function. For the reported layout the `has_no_monitor_to_left` and `has_no_monitor_to_right` correctly evaluated to false. But the `has_no_monitor_above` wrongly evaluated to `false`, so it was not possible to set structs. This caused the window to go under the wingpanel.

This was no problem for all 2 monitor layouts as the struts could always be set from the left or right side.

 With this branch, the wingpanel should now reserve the correct space for all layouts up to 3 monitors.

andreasfelix:
I made the code more concise by moving the logic into a single loop. Using 3 different function `has_no_monitor_above`, `has_no_monitor_to_left` and `has_no_monitor_to_right` was unnecessary repetitive and was the reason I introduced the bug in the first place. This should be more maintainable and easier to understand.

andreasfelix:
Isn't there a tool to create "virtual" monitors? Maybe this Xephyr program, which you use to test the greeter, is capable to do this? If so, this could also be useful to test other multi-monitor related issues on a single monitor...

quequotion:
So, not sure if you're interested in dealing with this, but resulting from commit 30d0757d2d140e7da53913b1a9861da0598549a2, when using wingpanel in openbox, windows cannot maximize larger than the right half of the screen (ie, there appears to be a strut occupying 50% of the screen from the left edge).

I built each version before and after to test; the fix for this bug causes another. I know openbox is outside of your purview, but I think there must be a better way to fix that issue that doesn't cause this one.

andreasfelix:
I am able to reproduce this. To me it seems like OpenBox is ignoring the `left_start_y` and `left_end_y`. because If I log them they have the correct values: 0 and 29 on a single FHD Monitor. 
I assume only the half monitor is blocked because of some maximal struct threshold in OpenBox. But this is just speculation, I did not look at the source code.

A workaround for OpenBox is to swap the order in the if statement:

https://github.com/andreasfelix/wingpanel/blob/f5e048321c69ffec0278b8cbb623928afd6c4f69/src/PanelWindow.vala#L172-L186

So it is the struts are set from the top if possible. This would make wingpanel work correctly on a single monitor on OpenBox and some other layouts, but for example not for a vertical dual monitor setup. This should have no effect on Gala.

PR #327, opened 2020-06-15
Workaround for OpenBox

andreasfelix:
See (https://github.com/elementary/wingpanel/pull/295#issuecomment-644239633) I just swapped the order of the if statement, this should have no effect on gala.

This should make wingpanel work correctly for some Multi-Monitor layouts in OpenBox, but e.g. not for vertical layouts, where it is necessary to set the struts from the left or right side.


quequotion:
This works for me, by the way.

jeremypw:
CLosing as not reviewed/actioned after 2+yrs and conflicts

PR #335, opened 2020-07-13
Remove deprecations before 3.0.0

danrabbit:


jeremypw (approved):
I guess they gotta go sometime ...
Still builds and runs OK.

PR #346, opened 2020-09-15
Wingpanel → Application: Remove get instance method

### LINKED ISSUE #344: Wingpanel dissapears and reapears
Hi, this issue keeps coming back everytime I interact with the panel : every time I click an applet, it would randomly disappear and reappear. Do you know how I could fix this ?

danrabbit:
* Move Wingpanel.vala → Application.vala
* Inline namespace
* main at the bottom like our other apps
* No get instance property


jeremypw:
Best to rename the file in a separate commit otherwise difficult to see what else has changed.

jeremypw (approved):
As far as I can tell, the instance method of Application is not used internally other than in main () and is not accessible externally.  The branch compiled and installed OK and the indicators I have continued to work as expected.

PR #349, opened 2020-09-20
Panel: simplify menubars and code style

danrabbit:
Menubar border width is 0 by default, so it doesn't seem like we need a whole class to set a single property

* Gobject style construct
* code style
* bump copyright


tintou (approved):

jeremypw (approved):
Makes sense. No regressions found.

PR #356, opened 2020-10-19
PanelWindow: Support getting monitor dimensions under wayland

davidmhewitt:
Currently if you try to run wingpanel under Wayland, it ends up as the minimum size because `get_primary_monitor` returns `null` on Wayland.

We'll need to add some sort of interface for getting the primary monitor in the future, but this at least gets wingpanel to the correct size (and hence usable) on single-monitor machines for now. It might do the wrong thing on multi monitor machines depending on the order.


pushqrdx:
may i ask how are you launching the panel under wayland? it's falling back to xwayland here

src/Utils.vala
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ */
+
+namespace Wingpanel.Utils {
+    public static bool is_wayland () {
+        return !(Gdk.Display.get_default () is Gdk.X11.Display);
```
danrabbit, line 22: Do we need a whole utils class for this one-liner?

danrabbit (approved):

Issue #366, opened 2021-01-24
App indicators appear in between system indicators


pongloongyeat:
## What Happened

App indicators appear in between system indicators.
![image](https://user-images.githubusercontent.com/31680656/105627961-ecfaaa80-5e74-11eb-9a98-b5d3cf3f0c6b.png)

## Expected Behavior

Although app indicators aren't supported, there should be at least some check to ensure that system indicators appear on the far right with any other custom indicator appearing to the left of system indicators.

## Steps to Reproduce

1. Turn on night light.
2. Install random app indicator ([wingpanel-indicator-template](github.com/pongloongyeat/wingpanel-indicator-template) for example).

danrabbit:
There's no way to ensure this since as far as the panel is concerned there's no such distinction between types of indicators. There is only indicators

davidmhewitt:
Sounds like we just need to add the nightlight indicator here, along with any other new ones we've missed:
https://github.com/elementary/wingpanel/blob/ef0908ea25f541f7e4944819e473f96d972ef048/src/Services/IndicatorSorter.vala#L33-L45

pongloongyeat:
Huh never knew there was a printer indicator and I do wonder what the sync indicator is :open_mouth:

Here's the list of un-added indicators for your convenience:

- wingpanel-indicator-datetime
- wingpanel-indicator-nightlight
- wingpanel-indicator-privacy
- wingpanel-indicator-a11y

davidmhewitt:
This is now resolved, thanks again @pongloongyeat !

PR #367, opened 2021-01-24
Add codenames for missing indicators

pongloongyeat:
This adds the following three constants:

```vala
public const string NIGHT_LIGHT = "nightlight";
public const string PRIVACY = "privacy";
public const string ACCESSIBILITY = "a11y";
```

Let me know if the indicator sorting as in #366 should be in this PR. I've left it out of this PR so the design team can add it in instead.

### DIFF
--- a/lib/Indicator.vala
+++ b/lib/Indicator.vala
@@ -29,6 +29,9 @@ public abstract class Wingpanel.Indicator : GLib.Object {
     public const string PRINTER = "printer";
     public const string BLUETOOTH = "bluetooth";
     public const string KEYBOARD = "keyboard";
+    public const string NIGHT_LIGHT = "nightlight";
+    public const string PRIVACY = "privacy";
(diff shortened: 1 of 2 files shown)


davidmhewitt:
I think it would be better to add the code for the ordering in this PR too.

It looks like indicators are sorted alphabetically if they are unknown. So to keep the sorting the same as it is now, you can set the order of the indicators as `a11y`, `nightlight`, `privacy` before all of the already existing indicators.

pongloongyeat:
@davidmhewitt it's done.

davidmhewitt (approved):
I'm going to approve and merge this as it will result in the same ordering of the native indicators. We can then iterate on that ordering if desired when all of the branches for the individual indicators have been merged.

Thank you for your PR @pongloongyeat

pongloongyeat:
Glad to be of help!

Issue #375, opened 2021-02-12
library name unnecessarily verbose; soversion and library file names inconsistent


decathorpe:
With wingpanel 2.x, the installed library file names are (directory layout might differ on ubuntu based systems):

- /usr/lib64/libwingpanel-2.0.so.**0**
- /usr/lib64/libwingpanel-2.0.so.**0**.2.0

This is, the versioned library **name** (-2.0 suffix) aside, consistent, and matches expectations.

When building the current git master branch (a5252c0) the resulting file names after installing are:

- /usr/lib64/libwingpanel-3.0.so.**0**
- /usr/lib64/libwingpanel-3.0.so.**3**.0.0

This seems to happen because soversion is set to `0` and lib_version to `3.0.0`. Almost always, the soversion is the leading digit of the "library version", which makes this very weird and inconsistent as of wingpanel 3.

---

I understand that an soversion of 0 can be kept since you're always changing the library **name** (not only the version) when making incompatible changes, but I don't think that's actually necessary? "Normal" libraries would do something like this:

    /usr/lib64/libwingpanel.so.3
    /usr/lib64/libwingpanel.so.3.0.0

(Dropping the redundant -3.0 version suffix from the library name, bumping soversion to 3. This would also allow you to just drop the -3.0 suffix from pkgconfig, vala, vapi, includedir, etc., making transitions to new major wingpanel versions much easier - just requiring a rebuild of a dependent package instead of necessitating actual code changes.)

PR #377, opened 2021-02-16
Change library name to "wingpanel" and use the right soversion

tintou:
This makes it easier to change the major version number afterwards.

Fixes https://github.com/elementary/wingpanel/issues/375


danrabbit (approved):
Yay RDNN

PR #379, opened 2021-02-17
Update packaging to new naming of the library and binary

tintou:


danrabbit (approved):
Build seems to succeed locally. LFG :rocket:

PR #391, opened 2021-04-04
Make labels not bold

jwinnie:
In my opinion, it looks quite a bit better and more readable, especially with the new font:

wingpanel


cassidyjames:
Gonna be a :-1: from me and it sounds like that's the consensus from @elementary/ux as well. It make it harder to read with no real benefit.

jwinnie:
Awww man ok

PR #393, opened 2021-04-12
Suppress tooltips when popover is open

### LINKED ISSUE #392: Suppress indicator tooltip when popover is visible
## What Happened

When the popover of an indicator is open/visible, the newly-implemented tooltips can overlap the popover, partially obscuring it from view.

## Expected Behavior

Suppress the tooltip while the popover is visible.

## Steps to Reproduce

1. Click on an indicator to make its popover visible.
2. Hover of the indicator display widget to trigger the tooltip.
3. Observe the issue where the popover is partially obscured from view.

### Logs

(none)

## Platform Information

Latest Odin early access installation.

vjr:
Fixes #392 

@pongloongyeat  I am unable to assign reviewers - would you mind reviewing or requesting appropriate developer to look at this?

### DIFF
--- a/src/Services/PopoverManager.vala
+++ b/src/Services/PopoverManager.vala
@@ -23,12 +23,10 @@ public class Wingpanel.Services.PopoverManager : Object {
     private bool grabbed = false; // whether the wingpanel grabbed focus
     private bool mousing = false;
     private bool is_popover_open = false;
-    private string tooltip_markup;
 
     private Gee.HashMap<string, Wingpanel.Widgets.IndicatorEntry> registered_indicators;
     private Wingpanel.Widgets.IndicatorPopover popover;
     private Wingpanel.Widgets.IndicatorEntry? _current_indicator = null;
-    private Wingpanel.Widgets.IndicatorEntry? _previous_indicator = null;
     public Wingpanel.Widgets.IndicatorEntry? current_indicator {
         get {
             return _current_indicator;
@@ -38,8 +36,6 @@ public class Wingpanel.Services.PopoverManager : Object {
             if (value == null && _current_indicator == null) {
                 return;
             }
-
-            _previous_indicator = _current_indicator;
 
             if (_current_indicator == null && value != null) { // First open
                 _current_indicator = value;
@@ -57,6 +53,7 @@ public class Wingpanel.Services.PopoverManager : Object {
             if (_current_indicator != null) {
                 popover.set_content (_current_indicator.indicator_widget);
                 popover.relative_to = _current_indicator;
+                _current_indicator.display_widget.has_tooltip = false;
                 owner.set_expanded (true);
                 make_modal (popover, true);
                 owner.present ();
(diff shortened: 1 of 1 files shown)


vjr:
I guess I should use the `unowned` keyword for the display widget vars. Will update later in the day.

vjr:
At least one other issue with this PR is it assumes the tooltip is set in the markup property and does not save/restore the plain text tooltip property - what if there are third party indicators that do this?

Is there any way to distinguish which is the active property set by an indicator?

jeremypw (commented):
How about connecting to the `query-tooltip` signal and return `false` to suppress the tooltip? Then it would not matter what kind of tooltip it is.

vjr:
> How about connecting to the `query-tooltip` signal and return `false` to suppress the tooltip? Then it would not matter what kind of tooltip it is.

Oh I see, thank you, let me look into that (might take me a few days to revisit this PR), it would simplify things a lot I guess!

vjr:
Hi @jeremypw looks like you already have a branch with the simplified approach here -> https://github.com/elementary/wingpanel/tree/suppress-tooltips which looks good to me so I guess you can go ahead and have it merged 👍🏻

vjr:
@jeremypw  I guess around line 30 in your branch the `_previous_indicator` declaration can be removed.

jeremypw:
@vjr Huh - I had completely forgotten about that branch! Not sure why I did not submit it.  I'll have a look at it again and submit. Thanks.

jeremypw:
@vjr It looks like I did push that branch as a PR to this one (https://github.com/vjr/wingpanel/pull/1).  So I have now removed the unused code. Its up to you to merge it into this one.

vjr:
> 
> 
> @vjr It looks like I did push that branch as a PR to this one ([vjr#1](https://github.com/vjr/wingpanel/pull/1)). So I have now removed the unused code. Its up to you to merge it into this one.

Done!

jeremypw (approved):
I am OK with this now, but as I contributed some code maybe should wait a bit to see if there are further reviews.

marbetschar:
Not tested this, but instead of introducing an additional variable (`is_popover_open`) isn't there already an instance variable of the [Gtk.Popover](https://valadoc.org/gtk+-3.0/Gtk.Popover.html) itself which does reflect the open/close state? Maybe [`Gtk.Widget.visible`](https://valadoc.org/gtk+-3.0/Gtk.Widget.visible.html)?

vjr:
> Not tested this, but instead of introducing an additional variable (`is_popover_open`) isn't there already an instance variable of the [Gtk.Popover](https://valadoc.org/gtk+-3.0/Gtk.Popover.html) itself which does reflect the open/close state? Maybe [`Gtk.Widget.visible`](https://valadoc.org/gtk+-3.0/Gtk.Widget.visible.html)?

Thanks for the comment! Yes, let me take a look and see if it works and I'll update this PR if it does...

vjr:
@marbetschar looks like the `visible` property works!

@jeremypw i had to make some minor changes in addition to using the `visible` property because the tooltip was not being restored in some cases (likely when i was switching between adjacent popovers while they were open) - and i noticed you used `.indicator_widget` in some places and `.display_widget` in others; was there a reason for that? i've just used `.display_widget` everywhere...

in any case, the PR is whittled down to just a 3-line diff now 👍🏻

src/Services/PopoverManager.vala
```diff
@@ -60,16 +60,9 @@ public class Wingpanel.Services.PopoverManager : Object {
...
                 _current_indicator.base_indicator.opened ();
-                is_popover_open = true;
-                _current_indicator.display_widget.notify["has-tooltip"].connect (() => {
-                    if (is_popover_open) {
-                        _current_indicator.display_widget.has_tooltip = false;
-                    }
-                });
             } else {
-                ((Wingpanel.Widgets.IndicatorEntry)popover.get_relative_to ()).indicator_widget.has_tooltip = true;
+                ((Wingpanel.Widgets.IndicatorEntry)popover.get_relative_to ()).display_widget.has_tooltip = true;
```
vjr, line 64: @jeremypw this is where I replaced the `.indicator_widget` with `.display_widget` FYI to take a look if it's OK or not?

jeremypw: You are right, it should be display_widget as that is what the tooltip is attached to.  I must got confused with the naming.

jeremypw:
Other than a minor quibble about the `suppress_tooltip ()` function this is looking good. Thanks.

vjr:
> Other than a minor quibble about the `suppress_tooltip ()` function this is looking good. Thanks.

There is already a similar existing private function called `make_modal ()` in this file with a boolean flag that inverts the behaviour, but do you have a suggestion for a better name than `suppress_tooltip ()` ? Maybe `update_has_tooltip (xxx, bool enable = true)` ?

src/Services/PopoverManager.vala
```diff
@@ -165,6 +168,12 @@ public class Wingpanel.Services.PopoverManager : Object {
         return current_indicator != null && current_indicator.base_indicator.code_name == entry.base_indicator.code_name;
     }
 
+    private void suppress_tooltip (Gtk.Widget display_widget, bool suppress = true) {
```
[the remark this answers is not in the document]
vjr: See my comment about the function name -> https://github.com/elementary/wingpanel/pull/393#issuecomment-870670363

The reason I added this function to do the null check is because in https://github.com/elementary/wingpanel/blob/master/src/Widgets/IndicatorEntry.vala#L54 it appears the `display_widget` could be left null; so no harm in doing a null check here at worst.

jeremypw:
If the display widget can validly be `null` then it ought to be defined as nullable in the code and it isn't.  I cannot see a valid case for having an indicator without a display widget tbh. I suppose 3rd party indicators might do something weird but I am not sure we still support those.  However, I guess it does no harm to be on the safe side and keep this check.  In which case `update_has_tooltip ()` is a reasonable name.

vjr:
> If the display widget can validly be `null` then it ought to be defined as nullable in the code and it isn't. I cannot see a valid case for having an indicator without a display widget tbh. I suppose 3rd party indicators might do something weird but I am not sure we still support those. However, I guess it does no harm to be on the safe side and keep this check. In which case `update_has_tooltip ()` is a reasonable name.

@jeremypw done!

jeremypw (approved):
Let's go!

PR #413, opened 2021-08-31
Add initial support for Mutter 41

tintou:


decathorpe:
This does not compile with some problem in the new vapi files ... maybe something that was supposed to be a symlink is a copy instead? There's hundreds of errors like this one:

```
libmutter.vapi:11.3-11.42: error: `Meta.Prefs' already contains a definition for `change_workspace_name'
		public static void change_workspace_name (int i, string name);
		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
libmutter-9.vapi:11.3-11.42: note: previous definition of `change_workspace_name' was here
		public static void change_workspace_name (int i, string name);
		^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

tintou:
@decathorpe You need Gala built against libmutter-9

decathorpe:
I do have that. Otherwise, the wingpanel build doesn't even get that far because it can't install both mutter-3.38-devel and mutter-41-devel into the build environment.

tintou:
That's likely because of https://github.com/elementary/gala/pull/1238

decathorpe:
I'll try that later. thanks!

decathorpe:
Indeed, a resubmitted build against the very latest gala did succeed. I have not been able to check if it works yet as I'm not at my workstation right now, but I don't think that there should be problems.

decathorpe (approved):

PR #428, opened 2021-12-18
Application: Only load dark style preference in session

danrabbit:
This is kind of a hotfix for startup delays on the greeter. I'm not sure what the proper long-term solution should be, but this resolves it in a simple way


davidmhewitt (approved):
Makes sense!

PR #445, opened 2022-04-07
Fallback to maximized style if Gala bus is unavailable

davidmhewitt:
Currently, if you run wingpanel on a compositor that doesn't have the Gala DBus interface for checking the wallpaper colour, it defaults to a fully transparent light style which may not be ideal depending on the wallpaper.

It would be better to just have to solid black maximized panel in cases where the Gala DBus interface isn't available. This would either be because the compositor isn't Gala and doesn't implement the interface, or because Gala crashed (and you've got bigger issues than the colour of your panel)

### DIFF
--- a/src/Services/BackgroundManager.vala
+++ b/src/Services/BackgroundManager.vala
@@ -76,7 +76,12 @@ namespace Wingpanel.Services {
 
             Bus.watch_name (BusType.SESSION, DBUS_NAME, BusNameWatcherFlags.NONE,
                 () => connect_dbus (),
-                () => bus = null);
+                () => {
+                    bus = null;
+                    // If the Gala bus is unavailable or vanishes, fall back to maximized style,
+                    // as this is most visible on all backgrounds
+                    background_state_changed (BackgroundState.MAXIMIZED, 0);
+                });
         }
 
         public void remember_window () {


danrabbit (approved):
Makes sense to me!

bobby285271:
When packaging the latest updates I have a quick try for this PR (https://github.com/NixOS/nixpkgs/pull/167698, https://github.com/NixOS/nixpkgs/commit/3b27254b41ca122f4fd4848e7a1446dd71b1eeec), and noticed that the greeter look wired for me after this change. Is there I should do downstream to resolve this?

屏幕截图 2022-04-08 19 06 31

davidmhewitt:
@bobby285271 Thanks for noticing this!

I guess this is an unintended effect of this PR. The greeter doesn't use Gala as a compositor, so is affected by this change.

It's just lucky (or intentional) that the default of wingpanel was previously the fully transparent style. So I guess we want to revert this PR, or implement a skeleton of the needed DBus interface in the greeter compositor, so it can control the colour of the panel itself, even if it is just hardcoded to transparent.

danrabbit:
@davidmhewitt I'd like to revisit having a blurred version of the wallpaper as the background so I'm in favor of implementing the dbus in the greeter compositor :)

PR #449, opened 2022-04-15
Styles: fix CSS syntax, remove extra class

danrabbit:
Some style fixes ahead of Gtk4 to reduce the diff there

* Fix CSS syntax error with `alpha`
* Remove old `window-dragging`
* Set a min-height to avoid divide by zero
* Don't see `menubar` because it doesn't seem like there's a reason to set it here and the class doesn't exist in Gtk4
* Remove hardcoded margin in IndicatorEntry

### DIFF
--- a/data/styles/application.css
+++ b/data/styles/application.css
@@ -8,12 +8,12 @@
     color: #fff;
     font-weight: bold;
     text-shadow:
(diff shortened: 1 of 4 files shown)


JoseExposito (approved):
Looks good :)

PR #470, opened 2022-10-13
build: Fix building with gala 6.3.2

bobby285271:
See https://github.com/elementary/gala/issues/1471

Since the error comes from gala.vapi, I guess the posix_dep can be placed next to gala_dep?


tintou:
Good catch, the issue is in Gala then

bobby285271:
Then I will close this? :laughing:

PR #485, opened 2023-06-22
Added install_deps.sh and edited README.md

deathbot159:
Hey, 
I have added `install_deps.sh` for installing dependencies required to build an application. Also, this script is configuring the build environment and adds `build.sh` and `install.sh` scripts for (in my opinion) better dev work.
I've changed README.md, because to deprecated `libmutter-2-dev` (replaced with `libmutter-10-dev` which is available), and added instructions for scripts too (replaces original methods).

I hope that should help a little with getting started for editing of app :)


jeremypw:
Is running `sudo apt build-dep io.elementary.wingpanel` not sufficient to install the dependencies?

danirabbit:
Yeah gonna close this since there's already an apt-specific way to install deps automatically. Thanks for the idea though!

PR #486, opened 2023-07-16
Fix indicator re-animating when an icon is removed/added

### LINKED ISSUE #106: Indicators re-animate when an icon is removed
1. Disable bluetooth and notifications
2. Have nightlight running, then disabled it (not snooze, disable)
3. When the nightlight icon is animated out of the panel, notice that the bluetooth and notifications icons re-slash

It appears that when an icon is removed from the panel, all of the icons (maybe even the whole menubar?) are destroyed and recreated. This causes these icons to re-animate.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/57641206-bluetooth-and-notifications-indicators-re-animate-when-nightlight-icon-is-removed?utm_campaign=plugin&utm_content=tracker%2F58844522&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F58844522&utm_medium=issues&utm_source=github).
</bountysource-plugin>

lenemter:
Fixes #106

### DIFF
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -21,6 +21,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
     public Services.PopoverManager popover_manager;
 
     private Widgets.Panel panel;
+    private Gtk.EventControllerKey key_controller; // For keeping in memory
     private int monitor_number;
     private int monitor_width;
     private int monitor_height;
@@ -78,6 +79,9 @@ public class Wingpanel.PanelWindow : Gtk.Window {
         application.set_accels_for_action ("app.cycle-back", {"<Control><Shift>Tab"});
 
         add (panel);
+
+        key_controller = new Gtk.EventControllerKey (this);
+        key_controller.key_pressed.connect (on_key_pressed);
     }
 
     private bool animation_step () {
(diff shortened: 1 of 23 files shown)


jeremypw (changes requested):
The problem with this is that the order of the items in wingpanel panel varies when it is restarted - e.g. the bluetooth indicator ends up on the far  right, where the session indicator should be. And the numlock indicator inserts into the middle of the indicators.

The indicators should remain in a consistent order regardless of disabling them or turning numlock on/off

It does fix the issue though.

jeremypw (approved):
Fixes the issue.

It is possible to avoid keeping a second list  by iterating over the first list and counting visible items before the target item but it doesn't save much code.

lenemter:
@jeremypw I removed second list because it made code pretty hard to understand. Can you check if it still works fine on your computer please?

jeremypw (commented):
Fixes the issue for me.

There is a relatively short single-use function which generally should be avoided, though.

jeremypw (approved):
All good - lets go!

PR #487, opened 2023-07-16
Close popover on Esc

### LINKED ISSUE #283: Esc key to close wingpanel and return to previous window
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
**Is your feature request related to a problem? Please describe.**

Add an option to use the `Esc` key to close the wingpanel after clicking on it and return to previous window. For example, if I have Files open and I click on the WiFi (Network settings), the focus shifts to WiFi settings of course and then we should be able to use `Esc` key to close that and return focus to Files.

**Describe the solution you'd like**

Add `Esc` key functionality to return focus to previously opened window

lenemter:
Fixes #283

### DIFF
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -21,6 +21,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
     public Services.PopoverManager popover_manager;
 
     private Widgets.Panel panel;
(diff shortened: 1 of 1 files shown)


jeremypw:
Looks like this might be ported to Gtk4 some time so maybe implement this using `Gtk.EventControllerKey`?

PR #491, opened 2023-08-09
PanelWindow: animate in with Gtk.Revealer

danirabbit:
Replace window moving animation stuff with a Gtk.Revealer


jeremypw (approved):
Code looks good and no regressions found.

PR #494, opened 2023-08-25
3.0.4 release notes

lenemter:
Contains a bugfix and a small new feature. It would be nice to release a new version before introducing significant changes (#492 or #481)


lenemter:
Lint CI fails o_O

PR #497, opened 2023-08-25
Application: hard code elementary styles

danirabbit:
Make sure people can't break styles even if they try to change the system stylesheet or icons

### DIFF
--- a/data/wingpanel.appdata.xml.in
+++ b/data/wingpanel.appdata.xml.in
@@ -10,11 +10,13 @@
   <description>
     <p>The extensible top panel for Pantheon. Wingpanel is an empty container that accepts indicators as extensions, including the applications menu.</p>
   </description>
+
   <releases>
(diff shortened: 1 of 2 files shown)


zeebok (approved):

DimasKunJ:
how do i use another theme?. i don't really like the default theme. I can't change it and also can't downgrade.

danirabbit:
@DimasKunJ We don't support changing the system style sheet. It's required to avoid buggy behavior

rdanter:
I love elementaryOS, but some of the icons are not good nor even consistent with others. Could you at least make it possible to change icons? Perhaps with an "expert" mode or something to prevent people doing it accidentality?

danirabbit:
@rdanter if you have an issue with one of the icons please report it at https://github.com/elementary/icons

If you'd like to request a new feature in the panel please open a new issue report here.

Closed pull requests are not a great place to try to have discussion 

Thanks!

PR #499, opened 2023-09-02
Use dark style in greeter session

lenemter:
Complementary to https://github.com/elementary/greeter/pull/684


danirabbit (approved):

lenemter:
@davidmhewitt https://github.com/elementary/greeter/pull/684 both launches portals in the greeter session and syncs some dark style preference from selected user's AccountsService to lightdm's one.

PR #518, opened 2024-02-08
Fix typo on log message

desiderantes:


danirabbit (approved):
Thanks for the typo fix :)

PR #525, opened 2024-03-09
PopoverManager: simplify `make_modal`

lenemter:


jeremypw:
The code is OK but I don't really understand the purpose of this function tbh - can't the popover `modal` property just be set?  It will have to be rewritten for Gtk4 anyway as there is no grabbing there.

PR #526, opened 2024-03-09
PanelWindow: simplify `set_expanded`

lenemter:


jeremypw (approved):
Code looks equivalent.

PR #493, opened 2023-08-09
Clean up CSS and fix issues with shadows and struts

danirabbit:
Fixes #408 

* Use a single `Application.css` file instead of separate `application.css` and `panel.css` (In Gtk4 we can use Granite.init to load this automatically)
* Set `panel` css name once instead of setting style class twice nested within itself
* Update easing func
* Remove a constants class for strings that are only used once
* Fix drop shadow not rendering when translucent styles change (for example from a solid color to a busy wallpaper)

## BEFORE
Screenshot from 2023-12-06 10 43 53
Screenshot from 2023-12-06 10 50 14

## AFTER
Screenshot from 2023-12-06 10 42 33
Screenshot from 2023-12-06 10 49 27

PR #527, opened 2024-04-03
wayland: Fix wingpanel breaking everything

### LINKED ISSUE #515: wayland: Opening multitasking view breaks gala
### What Happened?

No idea what's going on here but opening the multitasking view causes behavior like when gala crashes on X. One can still somewhat interact with windows but resizing them does weird stuff and multitasking view and window switcher are broken. What's especially weird is that just switching to another tty and back fixes the whole thing until the multitasking view is opened again. 

I can't remember this happening when gala started to work on wayland but I tested commits up to the wayland fixes and it was the same everywhere. So maybe I don't remember correctly or another component is the cause?

### Steps to Reproduce

1. Launch wayland session
2. Open multitasking view

### Expected Behavior

Dont break

### OS Version

7.x (Horus)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #515

I've got no idea what's going on here but this small line of code broke pretty much everything on wayland (See the linked issue and its comment). The only workaround currently is killing wingpanel first thing you do when getting into the wayland session before touching anything else.

@danirabbit this was added recently with #493. Everything still works fine for me without it but I didn't test a lot so maybe you can give some background on why it was added?


danirabbit:
Ah yeah this has to do with setting the panel size/struts when styles change between transparent and translucent. It might be an X specific issue

src/PanelWindow.vala
```diff
@@ -81,8 +81,6 @@ public class Wingpanel.PanelWindow : Gtk.Window {
 
         key_controller = new Gtk.EventControllerKey (this);
         key_controller.key_pressed.connect (on_key_pressed);
-
-        panel.size_allocate.connect (update_panel_dimensions);
```
danirabbit, line 85: Maybe instead for now we do this?
```
if (!Utils.is_wayland ()) {
    panel.size_allocate.connect (update_panel_dimensions);
}
```

leolost2605: Sure that works, sorry kinda forgot about this PR :grimacing:

danirabbit (approved):

PR #533, opened 2024-05-24
Use wayland protocol for positioning

leolost2605:
TODO:

- [x] Request focus on opening applications menu
- [x] ~Maybe:~ Get primary monitor (we can just use the one we are on because gala makes sure we are always on the primary one)


danirabbit (approved):
LFG :rocket:

PR #536, opened 2024-05-28
Remove autostart file

leolost2605:
We are started by gala anyways and this will interfere with our stuff if gnome-session manages to start wingpanel before gala.

IMPORTANT: This will not be removed by an update (see elementary/session-settings#44) which means the wayland experience might be sub-par if not manually removed, I guess we need the post-install as with what we did with mail I guess?


danirabbit (approved):
This makes sense, we're starting with Gala even on X11 right?

PR #538, opened 2024-05-28
wayland: Don't use the revealer

leolost2605:
Keep it always revealed on wayland since animating in is done by gala. Also we don't need the size_allocated safeguard anymore it just interferes with us having the proper dimensions.

See elementary/gala#1930


danirabbit (approved):
Let's go :rocket:

PR #541, opened 2024-06-04
Don't expand window for popovers on Wayland

danirabbit:
Expanding the window is a workaround for X11 not being able to draw popovers outside of the window. This let's popovers go outside of the window and makes sure they're positioned to the bottom


leolost2605:
Makes sense and would be nice so that we can clean up even some of the protocol stuff but it seems that now only the middle indicator (date and time) automatically closes when I click outside :shrug:

leolost2605:
Did some more testing and AFAICT this will probably only work on GTK4 since GTK3 doesn't do proper xdg grabs on wayland?

PR #543, opened 2024-07-10
GTK4 prep: Move from MenuBar/Item to Box

leolost2605:


danirabbit (approved):
Makes sense and works here!

PR #544, opened 2024-07-10
GTK4 prep: Use event controllers for open and scrubbing

leolost2605:
~Also remove the grab last focused window operation.~


leolost2605:
Just a quick note since I've seen some activity here: I think this PR doesn't work on X not 100% sure though. On that matter: Should we try our best to keep X compatability in GTK4 or should we rather do something similar to dock?

danirabbit:
@leolost2605 Yeah we have to keep a working X11 session until the new a11y stack is available. So I think it'd be better to work on getting things as prepped as possible without introducing any breakage for now

leolost2605:
This now also removes the grabbing the last focused window feature since that currently doesn't seem to work on both X and wayland?

zeebok:
@leolost2605 There are now conflicts with master to resolve

leolost2605:
Conflicts are resolved and this should now keep exactly the same behavior as current master

danirabbit (changes requested):
afaict scrubbing works as expected, everything seems to work okay, except the panel drag behavior doesn't work. It looks like you used GestureMultiPress there, should it be instead a [GestureDrag](https://valadoc.org/gtk+-3.0/Gtk.GestureDrag.html)?

leolost2605:
> afaict scrubbing works as expected, everything seems to work okay, except the panel drag behavior doesn't work in X11. It looks like you used GestureMultiPress there, should it be instead a [GestureDrag](https://valadoc.org/gtk+-3.0/Gtk.GestureDrag.html)?

It's also currently broken on master but I now found a fix that works in tandem with this so I fixed it here :)

danirabbit (approved):
Nice job! I can confirm this works in X11 and Wayland now :)

PR #558, opened 2024-08-12
Fix window drag

leolost2605:
Based on #556 with a slightly adjusted method from elementary/gala#2019. Still doesn't do touchscreen on x but on wayland both work

Closes #556 
Fixes https://github.com/elementary/gala/issues/2011
Fixes https://github.com/elementary/gala/issues/2009


lenemter (changes requested):
Clicking on the panel begins window grab even if you don't hold down the mouse button.

leolost2605:
Should be fixed now :)

lenemter:
This branch has issues with window focus. Sometimes drag doesn't start and wingpanel is focused instead.

Another issue: lets say you have 2 windows: browser (maximized, not focused) and Files (non-maximized, focued). In https://github.com/elementary/wingpanel/pull/556 clicking the panel would focus the browser. In this merge request it works very unpredictably. If you click a bunch of times on the panel focus goes like Files -> Panel -> Browser -> Panel.

I think the way we should fix this is in gala listen to every event (`stage.captured_event`) and save last device to WindowManager and expose it over libgala, so both gala and wingpanel plugin can always get the latest device. If you have other ideas i'd be happy to hear them out.

leolost2605:
> This branch has issues with window focus. Sometimes drag doesn't start and wingpanel is focused instead.

This is fixed by the latest commit so that wingpanel never takes focus.

> Another issue: lets say you have 2 windows: browser (maximized, not focused) and Files (non-maximized, focued). In https://github.com/elementary/wingpanel/pull/556 clicking the panel would focus the browser. In this merge request it works very unpredictably. If you click a bunch of times on the panel focus goes like Files -> Panel -> Browser -> Panel.

Should it focus the browser or should it do nothing? On the latest it (should) do nothing except of course if you drag then the maximized window gets focused and a drag begins.

The issue with the drag sometimes not starting is I think with the Gtk.GestureDrag not triggering if the range you travel is to short so I switched back to `button_press_event` for now.

leolost2605:
> I think the way we should fix this is in gala listen to every event (stage.captured_event) and save last device to WindowManager and expose it over libgala, so both gala and wingpanel plugin can always get the latest device. If you have other ideas i'd be happy to hear them out.

The reason why I decided against that at least in the window menu case is that the last device doesn't have to be the device you want to move with, i.e. you can have something like:

open window menu -> with the touchscreen tap move -> move the pointer via touchpad

This wouldn't work if the grab op is started with the last device (the touchscreen used to click the menu item).

Besides hopefully everything should work now here since it mostly was issues on the wingpanel client side :crossed_fingers:

lenemter (approved):
Works flawlessly now. Good job!

PR #559, opened 2024-08-12
Fix fractional scaling

leolost2605:
Turns out we don't need to manually multiply anything :)

Otherwise we get very weird sizes


danirabbit (approved):
Looks good in low and HiDPI here

PR #563, opened 2024-08-15
PanelWindow: divide dimensions by scale on Wayland

### LINKED ISSUE #562: Most indicators missing on Wayland
### What Happened?

After upgrading my system, most indicators are missing on the Wayland session, only the calendar remains.

### Steps to Reproduce

1. Upgrade elementary
2. Login to the Wayland session
3. All the indicators are gone

### Expected Behavior

The indicators should be visible, particularly the power related ones

### OS Version

8.x (Early Access)

### Software Version

### Log Output

_No response_

### Hardware Info

Intel system with an AMD RX7600 GPU installed

danirabbit:
Fixes #562 

Test panel size on:
- [x] HiDPI with Wayland
- [x] LoDPI with Wayland
- [x] HiDPI with X11
- [x] LoDPI with X11

### DIFF
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -99,6 +99,12 @@ public class Wingpanel.PanelWindow : Gtk.Window {
 
         // We just use our monitor because Gala makes sure we are always on the primary one
         var monitor_dimensions = get_display ().get_monitor_at_window (get_window ()).get_geometry ();
+        if (Gdk.Display.get_default () is Gdk.Wayland.Display) {
+            monitor_dimensions.width /= get_scale_factor ();
+            monitor_dimensions.height /= get_scale_factor ();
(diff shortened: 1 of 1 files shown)


danirabbit:
Can't test on X11 because of #564 :(

leolost2605:
It seems this is only needed when fractional scaling is disabled (since it does logical layout instead of physical). Otherwise it will reintroduce the issue fixed in #559. Maybe we can do it depending on whether it's enabled?

danirabbit:
@leolost2605 ah is there a simple way to test for that?

leolost2605:
Hmm I dug a bit through gtk api but I only found stuff for GTK4 so I was kinda hoping somebody else knew something. I'll prepare a branch where we ask mutter via dbus :)

danirabbit:
Closing since you proposed a better branch

PR #566, opened 2024-08-16
Scale dimensions if we have a physical monitor layout

### LINKED ISSUE #562: Most indicators missing on Wayland

leolost2605:
Fixes #562
Closes #563

### DIFF
--- a/src/Services/DisplayConfig.vala
+++ b/src/Services/DisplayConfig.vala
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2024 elementary, Inc. (https://elementary.io)
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+
+[DBus (name = "org.gnome.Mutter.DisplayConfig")]
+public interface Wingpanel.Services.DisplayConfig : Object {
+    private static bool? _is_logical_layout = null;
+    private static DisplayConfig? proxy = null;
+
+    public static bool is_logical_layout () {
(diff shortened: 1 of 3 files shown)


danirabbit (changes requested):
With this branch I can confirm the panel size is correct with:

:heavy_check_mark: Wayland on HiDPI
:heavy_check_mark: Wayland on LoDPI
:heavy_check_mark: X11 on LoDPI

but

:x: X11 on HiDPI I get small panel

src/Services/DisplayConfig.vala
```diff
@@ -0,0 +1,94 @@
...
+        } catch (Error e) {
+            critical (e.message);
+            _is_logical_layout = true;
+            return;
+        }
+
+        update ();
+    }
+
+    private static void update () {
```
danirabbit, line 32: Should this return `true` early if we're on X11?

danirabbit (approved):
Gonna merge this because it's better than the current situation and panel is blocking other things

PR #567, opened 2024-08-19
Switch workspace when scrolling on the panel

### LINKED ISSUE #524: Mouse wheel scroll when hovering wingpanel switches workspaces
### Problem

Switching workspaces has long been a great power user feature, but isn't currently super well-exposed. Keyboard-heavy navigators can bind shortcuts to switch workspaces, but if your hand is already on the mouse, this proposal would frequently be more convenient.
Binding screen corner actions to workspace switching has the usual potential downsides, whereas currently mousewheel does nothing when the cursor is over the wingpanel.

### Proposal

When hovering the mouse over the wingpanel, mousewheel up and down navigate to higher and lower number workspaces, respectively. That's about it, really.

This behavior wouldn't even need to be on by default to still be useful a certain % of users who might appreciate it.

### Prior Art (Optional)

Numerous Gnome Shell extensions over the years have provided this feature, eg https://github.com/sunwxg/gnome-shell-extension-panelScroll

leolost2605:
Fixes #524

### DIFF
--- a/schemas/io.elementary.desktop.wingpanel.gschema.xml
+++ b/schemas/io.elementary.desktop.wingpanel.gschema.xml
@@ -6,5 +6,10 @@
 			<summary>Sets if the panel uses transparency.</summary>
 			<description>Disable this to provide higher contrasts and make indicators better readable.</description>
 		</key>
+		<key type="b" name="scroll-to-switch-workspaces">
+			<default>false</default>
+			<summary>Sets if scrolling on the panel switches workspaces.</summary>
+			<description>Sets if scrolling on the panel switches workspaces.</description>
+		</key>
 	</schema>
 </schemalist>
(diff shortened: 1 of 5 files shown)


lenemter:
I think you forgot to add Services/WMDBus.vala file to git here :sweat_smile:

danirabbit (changes requested):
Hm, I can see the utility here but I think we probably need an off-by-default gsetting for this. I can already here the reports of folks triggering this by accident and thinking this is a bug since there's no visual indicator and there's nothing else workspace related in the panel

leolost2605:
> I think you forgot to add Services/WMDBus.vala file to git here 😅

Whoops added it :)

> Hm, I can see the utility here but I think we probably need an off-by-default gsetting for this. I can already here the reports of folks triggering this by accident and thinking this is a bug since there's no visual indicator and there's nothing else workspace related in the panel

Sure we can do that

leolost2605:
Should we expose it in settings?

leolost2605:
Proposed the setting at elementary/switchboard-plug-pantheon-shell#416 :)

leolost2605:
For a better experience I also opened elementary/wingpanel-indicator-power#264

danirabbit (approved):
Works for me! Cool feature :)

JPLeBreton:
Thanks so much for implementing this! ❤️

PR #572, opened 2024-08-24
Use new gala dbus name

lenemter:


zeebok (approved):
This looks fine. Will we need to update other repos because of this?

danirabbit:
Yeah dont merge this yet it'll break everything. Dbus hasn't changed in gala yet

PR #574, opened 2024-08-31
PopoverManager: Remove a lot of (hopefully) unnecessary stuff

leolost2605:
So I've tested this quite a bit and it still seems to work perfectly fine on both X and wayland.
The thing is a lot of stuff is very old so there isn't a github PR linked and unfortunately the commits also don't have any meaningful commit messages so I can't say for certain for what reason some of this stuff was introduced.
Too fully work this requires #544 otherwise scrubbing doesn't work.
The reason for this PR is that some of the things aren't possible in GTK4


danirabbit (approved):
This looks good to me. Everything seems to work. I tested scrubbing, scroll events, maximized window dragging, etc in both sessions.

Might just be leftovers from when we had a custom popover widget before Gtk even had popovers! It's an old codebase :sweat_smile:

PR #575, opened 2024-08-31
PopoverManager: Simplify owner button press recognition

leolost2605:
Use EventControllers


danirabbit (approved):
Wild how much code this saves and seems to work exactly the same!

PR #576, opened 2024-09-01
PanelWindow: Only multiply by scale factor on wayland

### LINKED ISSUE #568: Panel is too small on HiDPI + X11
### What Happened?

The panel is smaller than it should be when using HiDPI and X11 together

### Steps to Reproduce

1. Use a HiDPI X11 session without enabling experimental fractional scaling
2. See that the panel is too small

### Expected Behavior

Panel is the correct size in all sessions. Probably should have a special case handler for X11 session in `src/Services/DisplayConfig.vala`

### OS Version

8.x (Early Access)

### Software Version

Compiled from git

### Log Output

_No response_

### Hardware Info

_No response_

leolost2605:
Fixes #568 

Not sure why but this works :shrug:


tintou:
Yeah that was my conclusion previously :shrug:

tintou (approved):

PR #579, opened 2024-09-17
Port to GTK4

### LINKED ISSUE #655: Tooltips position does not correct
### What Happened?

When I open the quick settings and move my mouse over the shutdown, suspend, or system settings buttons, the little help boxes (tooltips) pop up on the left side of the screen. They should be appearing right next to the button I'm pointing at.

Image

### Steps to Reproduce

1. Open Quick Settings
2. Move the mouse over those buttons

### Expected Behavior

They should be appearing right next to the button I'm pointing at.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

### Log Output

```shell

```

### Hardware Info

Image

leolost2605:
Fixes #655
closes #542 

This should work now.

Todo:

- [ ] window drag doesn't work on X11 (i'd like to treat this as not blocking because I'm not quite sure what's going on here and kinda consider X11 only legacy anyways but idk)
- [x] The transparency doesn't fully work (probably a CSS issue idk)

Misc:
- [ ] We get a warning about gtk_is_ancestor receiving null everytime an indicator is opened but idk why and wouldn't consider it blocking i guess?


danirabbit:
I think probably the best way forward for development is I can create a new packaging branch based on this branch and then build that in daily so we can start building indicators against it etc

But I'm not sure what trying to upgrade this mid-cycle looks like, so I'd like to leave it unmerged for now until we have a plan around how to handle the packaging/upgrade.

danirabbit (changes requested):
Don't forget to update deps on README and .github :)

danirabbit:
There's a packaging branch now: https://github.com/elementary/wingpanel/tree/deb-packaging-gtk4

So we should have access to the library on OS 8 daily no problem to work with porting the other indicators when this builds: https://code.launchpad.net/~elementary-os/+recipe/wingpanel-8-daily

You might want to temporarily change the meson project name here to `io.elementary.wingpanel-8` or something so that you can install them both at the same time for testing? Otherwise the `io.elementary.wingpanel-8` package will refuse to install because of the file conflicts. But this package name will eventually be reverted anyways so that's optional I guess :woman_shrugging:

danirabbit:
Looks like we didn't properly update the library stuff so we could at least try to build indicators:
```
 trying to overwrite '/usr/include/wingpanel/wingpanel.h', which is also in package libwingpanel-dev 8.0.3+r719+pkg73~daily~ubu
ntu8.1
```

src/Widgets/IndicatorEntry.vala
```diff
@@ -17,7 +17,11 @@
  * Boston, MA 02110-1301 USA.
  */
 
-public class Wingpanel.Widgets.IndicatorEntry : Gtk.EventBox {
+public class Wingpanel.Widgets.IndicatorEntry : Gtk.Widget {
```
danirabbit, line 20: We should use Granite.Bin now I suppose

src/Widgets/Panel.vala
```diff
@@ -17,7 +17,7 @@
  * Boston, MA 02110-1301 USA.
  */
 
-public class Wingpanel.Widgets.Panel : Gtk.EventBox {
+public class Wingpanel.Widgets.Panel : Gtk.Widget {
```
danirabbit, line 20: Also bin here

src/Widgets/Panel.vala
```diff
@@ -104,13 +109,17 @@ public class Wingpanel.Widgets.Panel : Gtk.EventBox {
...
+
     private void begin_drag (double x, double y) {
-        var window = get_window ();
+        var window = get_root ().get_surface ();
         if (window == null) {
             return;
         }
 
-        window.get_display ().get_default_seat ().ungrab ();
+        //  window.get_display ().get_default_seat ().ungrab ();
```
danirabbit, line 122: Commented code?

danirabbit (changes requested):

leolost2605:
> Looks like we didn't properly update the library stuff so we could at least try to build indicators:
> trying to overwrite '/usr/include/wingpanel/wingpanel.h', which is also in package libwingpanel-dev 8.0.3+r719+pkg73~daily~ubu
ntu8.1

@danirabbit I did something in #605 lmk if that's right :)

danirabbit:
I wonder if we should add this widget from Tasks since we have to replace ModelButton in a few places:

```
public class PopoverMenuitem : Gtk.Button {
    public string text {
        set {
            child = new Granite.AccelLabel (value) {
                action_name = this.action_name
            };

            update_property (Gtk.AccessibleProperty.LABEL, value, -1);
        }
    }

    class construct {
        set_css_name ("modelbutton");
    }

    construct {
        accessible_role = MENU_ITEM;

        clicked.connect (() => {
            var popover = (Gtk.Popover) get_ancestor (typeof (Gtk.Popover));
            if (popover != null) {
                popover.popdown ();
            }
        });
    }
}
```

danirabbit:
Finally starting to test this! It looks like we've got some issues with
- [ ] Indicators losing focus
- [ ] The panel not being transparent when it should be. Changing the wallpaper sets it correctly but then if you switch workspaces for example it'll go back to being opaque

Screenshot from 2025-06-06 12 12 44

danirabbit:
Interestingly, I can't reproduce the issue with indicators closing without accepting clicks on my Starlite running at 1.5x scaling. So I wonder if it has to do with scaling somehow?

danirabbit (approved):

leolost2605:
> Finally starting to test this! It looks like we've got some issues with
> 
>     * [ ]  Indicators losing focus
> 
>     * [ ]  The panel not being transparent when it should be. Changing the wallpaper sets it correctly but then if you switch workspaces for example it'll go back to being opaque
> 
>     * [ ]  Clicking inside an indicator closes that indicator and clicks on buttons etc aren't accepted

@danirabbit can you still reproduce these? I can't reproduce any of them

danirabbit:
@leolost2605 im still experiencing some issues where the panel is always using the translucent style and never the transparent style, but it was working correctly for me at one point. So I'm thinking this is maybe a problem with my local installation. I'll file a new issue report if I can reproduce it on a clean install.

Thanks for fixing the issue with indicators not opening with the margins etc!

decathorpe:
It looks like this changed the library name to `libwingpanel-9` but (accidentally?) left the version-based soname  in place. So it results in a library called `libwingpanel-9.so.8.0.4` now, but would immediately bump to `libwingpanel-9.so.9.0.0` (a breaking change) when 9.0.0 is tagged. Looks like this should now be using a hard-coded (i.e. not version-based) soname instead so it doesn't accidentally change when 9.0.0 is released.

Issue #588, opened 2024-11-30
Less heavy "scroll to change desktop"


teamcons:
### Problem

Current way to switch quickly desktop using the mouse is by quickly scrolling panel

However it happens quite often that i scroll on it and... nothing. It requires more scrolling for it to trigger - making it feel quite heavy

the debounce is a bit strong as well. If i want to peek quickly on another desktop, i cant immediately scroll back, but have to wait.

### Proposal

More sensitive scrolling
and shorter debounce

### Prior Art (Optional)

the default of what i used in Gnuhme was okay? i think it was this one.
it had a timeout of 0 between scroll event but im not sure if that would fit everyone

https://github.com/timbertson/gnome-shell-scroll-workspaces

teamcons:
Currently testing for a less heavy default.

PR #590, opened 2024-12-04
Snappier "top panel workspace scroll"

teamcons:
elementary's default implementation felt very heavy - you have to scroll a lot before it picks up anything, and the experience of going immediately a few workspaces left or right feels frustrating
The icon is too slow and keyboard not always in reach, so i use that feature daily.

"1" and "2" allows for a few misinputs (such as when mousewheel is between two notches somehow), but also picks up quickly enough user intent to react and switch
From using it a few days, ive set on "1" as "2" sometimes ignored valid inputs still. (roughly on my mouse thats the amount of notches that can be scrolled without causing a switch- I am not sure if other mouses would be different)

you cannot "scroll" too hard and go more than one workspace because the panel ignores further scrolling until the new workspace is fully shown. So low value is not an issue like it can be in another interpretations ive used, where you scroll a little too hard and end up too far in the workspace "stream" - Here you still end up scrolling one after the other

EDIT: Eventual refinements:
if the workspace switch isnt triggered by a scroll, the panel seems to "remember" the misinput and not accept any future one until it switches again

https://github.com/elementary/wingpanel/issues/588

### DIFF
--- a/src/Widgets/Panel.vala
+++ b/src/Widgets/Panel.vala
@@ -98,7 +98,7 @@ public class Wingpanel.Widgets.Panel : Gtk.EventBox {
 
             current_scroll_delta += dx + dy;
 
-            if (current_scroll_delta.abs () > 3) { //TODO: Check whether 10 is good here.
+            if (current_scroll_delta.abs () > 2) { //TODO: Check whether 10 is good here.
                 current_scroll_delta = 0;
             }
         });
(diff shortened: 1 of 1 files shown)


teamcons:
summoning @elementary/desktop-developers

alainm23:
I like it, it's faster.

alainm23 (approved):

teamcons:
> I like it, it's faster.

pondering whether it wouldnt even be better to remove that limitation, and scroll anytime a scroll is detected *at all*

danirabbit (approved):
tested this with both a mouse and a touchpad and this works for me. Sorry for taking so long to merge!

PR #601, opened 2025-03-07
ISSUE-594 fix panel actual_height with scale factor

### LINKED ISSUE #594: Half of the Wingpanel overlaps application windows
### What Happened?

From fresh installation any of the application windows draged to the top of the screen or when you are in maximized mode, half of the wingpanel overlaps it.
Issue1
Issue2

### Steps to Reproduce

1. Fresh elementary OS 8 installation
2. Hardware Macbook Pro 13" mid 2014 

### Expected Behavior

Wingpanel should not overlap application windows, since part of the menu is cut. 

### OS Version

8.x (Circe)

### Session Type

Classic Session (X11, This is the default)

### Software Version

### Log Output

_No response_

### Hardware Info

Macbook air 13" Mid 2014 
Intel HD Graphics 5000 - Retina Display 
2560 by 1600 pixels

lagoshny:
Changed condition to define right actual panel height with scale factor to prevent the overlay panel to other window in full mode.

Fixes #594

### DIFF
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -159,7 +159,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
     }
 
     private int get_actual_height () {
-        if (!Services.DisplayConfig.is_logical_layout () || Gdk.Display.get_default () is Gdk.Wayland.Display) {
+        if (!Services.DisplayConfig.is_logical_layout ()) {
             return get_allocated_height () * get_scale_factor ();
         }
 
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -159,7 +159,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
     }
 
     private int get_actual_height () {
-        if (!Services.DisplayConfig.is_logical_layout () && Gdk.Display.get_default () is Gdk.Wayland.Display) {
+        if (!Services.DisplayConfig.is_logical_layout () || Gdk.Display.get_default () is Gdk.Wayland.Display) {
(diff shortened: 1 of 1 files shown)


leolost2605:
The thing is this breaks when fractional scaling is enabled on wayland :\

For me the current version works on wayland both with
- fractional scaling
- 'normal' scaling i.e. 100 and 200 only i.e. physical layout

although it does require a wingpanel restart after scaling changed which definitely is something that can be optimized.

What combination would require this change? (X, wayland fractional or wayland not fractional)
Btw fractional is enabled when you add `scale-monitor-framebuffers` to the org.gnome.mutter.experimental-features dconf key

lagoshny:
This is necessary for the X11 Display. 
This behavior was the default until the wayland condition was added.

src/PanelWindow.vala
```diff
@@ -159,7 +159,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
     }
 
     private int get_actual_height () {
-        if (!Services.DisplayConfig.is_logical_layout () && Gdk.Display.get_default () is Gdk.Wayland.Display) {
+        if (!Services.DisplayConfig.is_logical_layout () || Gdk.Display.get_default () is Gdk.Wayland.Display) {
```
leolost2605, line 162:
```suggestion
        if (!Services.DisplayConfig.is_logical_layout ()) {
```

leolost2605 (changes requested):
Ah ok then you should probably just remove the condition because it shouldn't be always true on wayland.

leolost2605 (approved):
LGTM thank you!

PR #602, opened 2025-03-09
PanelWindow: Update size on scale change

leolost2605:


danirabbit (approved):
Works as expected! Nice fix

rifat151:
netspeed indicator in wingpanel elementary os 8 like this and tell me the process how can i do it.
49436911-4a5caa80-f788-11e8-911c-26baf6f3d40d

PR #605, opened 2025-04-18
Rename lib to wingpanel-8

leolost2605:
I did here what granite did. I'm not to familiar with this stuff so let me know if something looks off :sweat_smile: 
Also I wonder do we even want this? I guess for porting it's useful to not overwrite old indicators but in the end we should probably just overwrite them? So basically revert this commit?

### DIFF
--- a/lib/meson.build
+++ b/lib/meson.build
@@ -19,30 +19,31 @@ libwingpanel_deps = [
     gtk_dep,
 ]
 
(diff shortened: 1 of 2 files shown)


danirabbit (approved):
Thanks for doing this! I think it's still probably a good idea to bump the library even if ultimately the panel is not parallel installable

PR #611, opened 2025-05-09
Use dark style in greeter session

lenemter:


leonardo-lemos (approved):
Works as expected, nice job!

PR #618, opened 2025-06-09
Use translucent style with fullscreen window

lenemter:
Goes with https://github.com/elementary/gala/pull/2429


danirabbit (changes requested):
I think instead of maximized style this should maybe use `TRANSLUCENT` since it has some nice shadows etc

danirabbit (approved):
Let's go

PR #625, opened 2025-07-06
Update mutter dependency in readme

peteruithoven:


danirabbit:
It looks like we currently support as far back as `libmutter-10-dev`

danirabbit (approved):

Issue #414, opened 2021-09-18
Window switch on wingpanel click


AleVul:
## What Happened
Have 2 maximized windows, not necessary of the same application (ex.: Files and Music).
Open one of wingpanel's widgets, does not matter which one. While it is open, click on Wingpanel.
Notice:

1. widget closed
2. background window now is in foreground

A similar behavior can be observed when only one of the windows is maximized. To trigger the effect the non maximized window must be in foreground first. 
## Expected Behavior
only close the opened widget

## Steps to Reproduce

1. Open 2 applications, not necessary same app 
2. maximize both of the windows
3. open any wingpanel widget
4. while the widget is open, click on wingpanel

### Logs
Not sure where to get them or which ones

## Platform Information
OS: elementaryOS 6
wingpanel version: 3.0.0+r496+pkg55~ubuntu6.1

jeremypw:
I can confirm this in the case of a non-maximized window on top of a maximised window (but not for two maximised windows).  I noticed also that is the active window is maximised then dragging on the wing panel unmaximizes the window and starts dragging it rather than the wingpanel.  This suggests a click on the wingpanel gets passed on to a maximized window and therefore may bring it into focus.

jeremypw:
This might be an issue in Gala not the wingpanel.

PR #631, opened 2025-07-28
Initiate window drag only after drag threshold is reached

lenemter:
Fixes https://github.com/elementary/wingpanel/issues/133
Fixes https://github.com/elementary/wingpanel/issues/414

### DIFF
--- a/data/wingpanel.metainfo.xml.in
+++ b/data/wingpanel.metainfo.xml.in
@@ -44,7 +44,9 @@
         </ul>
       </description>
       <issues>
+        <issue url="https://github.com/elementary/wingpanel/issues/133">Don't initiate focus/drag on click/release</issue>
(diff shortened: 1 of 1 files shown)


wingpanel-interface/FocusManager.vala
```diff
@@ -128,6 +117,21 @@ public class WingpanelInterface.FocusManager : Object {
...
+                        event.get_coords (out start_x, out start_y);
+
+                        return Clutter.EVENT_PROPAGATE;
+                    }
+
+                    float current_x, current_y;
+                    event.get_coords (out current_x, out current_y);
+
+                    var drag_threshold = Clutter.Settings.get_default ().dnd_drag_threshold;
+                    if ((start_x - current_x).abs () <= drag_threshold && (start_y - current_y).abs () <= drag_threshold) {
```
zeebok, line 131: Should this be `&&` or `||` ? I would imagine you want `||` for drags that are mostly along an axis, no?

Issue #608, opened 2025-04-24
Custom indicators to be placed to the left


teamcons:
### Problem

Currently new indicators seems only allowed to the right

devs may want to offer alternative menus, not for indicating state, but for apps or files
or think a specific custom indicator may offer a placement to the user

it also feels very unbalanced for people who like clutter, with all the clutter only to the right

### Proposal

indicator property to center either right or left - but not leave them to be leftmost or rightmost either

PR #640, opened 2025-10-20
Lib/Indicator: Add enum for indicator position

teamcons:
Fixes https://github.com/elementary/wingpanel/issues/608

Add a property for indicators for initial placement.
For Launcher and Datetime, an ugly workaround give them the correct positioning until the property is declared in their respective repos.

i am not sure where to put the enum definition where it doesnt get lost in code. Having its own file for three lines feels overkills.

 - [x] Test this
 - [ ] More elegant ugly workaround

### DIFF
--- a/lib/IndicatorManager.vala
+++ b/lib/IndicatorManager.vala
@@ -61,29 +61,15 @@ public class Wingpanel.IndicatorManager : GLib.Object {
      */
     public signal void indicator_removed (Wingpanel.Indicator indicator);
 
-    /**
-     * Place the files in /etc/wingpanel.d/ or ~/.config/wingpanel.d/
-     * default.forbidden, greeter.allowed or combinations of it.
-     */
-    private Gee.HashSet<string> forbidden_indicators;
-    private Gee.HashSet<string> allowed_indicators;
-
     [CCode (has_target = false)]
     private delegate Wingpanel.Indicator? RegisterPluginFunction (Module module, ServerType server_type);
 
     private Gee.HashMap<string, Wingpanel.Indicator> indicators;
-
     private FileMonitor? monitor = null;
-
-    private FileMonitor? root_restrictions_monitor = null;
-    private FileMonitor? user_restrictions_monitor = null;
-
     private ServerType server_type;
 
     private IndicatorManager () {
         indicators = new Gee.HashMap<string, Wingpanel.Indicator> ();
-        forbidden_indicators = new Gee.HashSet<string> ();
-        allowed_indicators = new Gee.HashSet<string> ();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


lib/Indicator.vala
```diff
@@ -40,6 +40,11 @@ public abstract class Wingpanel.Indicator : GLib.Object {
      */
     public string code_name { get; construct; }
 
+    /**
+     * Set the indicator positioning on the panel
```
ryonakano:
```suggestion
     * Sets the indicator positioning on the panel.
```

Tiny thing but other existing properties use the Third Person form and have the period so let's honor them.

lib/IndicatorPosition.vala
```diff
@@ -0,0 +1,24 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public
+ * License along with this program; if not, write to the
+ * Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+ * Boston, MA 02110-1301 USA.
+ */
+
+public enum Wingpanel.IndicatorPosition {
```
ryonakano: > i am not sure where to put the enum definition where it doesnt get lost in code. Having its own file for three lines feels overkills.

Maybe putting this into Indicator.vala is fine? I agree with having its own file feels overkill.

src/Widgets/Panel.vala
```diff
@@ -247,12 +247,21 @@ public class Wingpanel.Widgets.Panel : Gtk.EventBox {
     private void add_indicator (Indicator indicator) {
         var indicator_entry = new IndicatorEntry (indicator, popover_manager);
 
-        switch (indicator.code_name) {
-            case Indicator.APP_LAUNCHER:
+        // Workaround until both indicators get their own position prop
```
ryonakano:
```suggestion
        // TODO Workaround until both indicators get their own position prop
```

Please explicit TODO so that this comment won't be buried into other normal comments.

src/Widgets/Panel.vala
```diff
@@ -247,12 +247,21 @@ public class Wingpanel.Widgets.Panel : Gtk.EventBox {
...
+        switch (indicator.position) {
+            case IndicatorPosition.LEFT:
                 indicator_entry.set_transition_type (Gtk.RevealerTransitionType.SLIDE_RIGHT);
-                left_menubar.add (indicator_entry);
+                left_menubar.insert_sorted (indicator_entry);
                 break;
-            case Indicator.DATETIME:
+            case IndicatorPosition.CENTER:
                 indicator_entry.set_transition_type (Gtk.RevealerTransitionType.SLIDE_DOWN);
                 center_menubar.add (indicator_entry);
```
ryonakano: I'm not sure if we may leave this `add()` or should update to `insert_sorted()`. The previous code was `add()` because it's only DateTime indicator that is placed in the center. However, now we're allowing showing multiple indicators in the center, so I guess we need to revisit this.

I'm concerning that possibly the indicator order of CENTER could be different every time when the panel starts.

teamcons: Properly centering opens a lot of uneasy questions:
-Would it mean we do a IndicatorPosition.CENTER_LEFT and CENTER_RIGHT, and leave a pure center for the date?
-Would it mean we rewrite a whole lot of code for basically a fourth and fifth position ?
-Is it worth it?

Im aware i am doing a PR for a new functionality, but also it is directed at a niche hacky extension that relies on using an unsupported (for user software) packaging system

So when doing the code i was unsure whether to address this properly, or wait until there is actual demand for actually putting in the center in a more proper way. Or also only allow LEFT or RIGHT.
Tbh im fine with all. The easiest path out would be to remove CENTER position, id be fine with adding a CENTER_LEFT + CENTER_RIGHT but that would take time out of other eOS related project to implement.

teamcons:
I dont wanna contribute to elementary OS anymore

PR #683, opened 2026-03-24
PopoverManager: use GLib.Once

danirabbit:
Make popovermanager a singleton instead of passing around references to it

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -110,6 +110,8 @@ public class Wingpanel.Application : Gtk.Application {
         panel_window = new PanelWindow (this);
         panel_window.present ();
 
+        var popover_manager = Services.PopoverManager.get_default ();
+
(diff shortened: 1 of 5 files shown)


leolost2605 (commented):
IMO we shouldn't convert to a Singleton. There's a bunch of literature that Singletons are considered an anti pattern except for a few very specific cases (e.g. a log).
https://en.wikipedia.org/wiki/Singleton_pattern#Criticism
In my experience in a few personal projects, not using singletons leads to a better structure because you have to think about who uses what. Also IMO the code it saves here isn't really code that we have to try to reduce, since it's not really any logic. But that's just my opinion based on what I found and heard from others 🤷

danirabbit:
That's a fair criticism and maybe this can be cleaned up in a better way. I'll approach it with a different mindset. Thank you!

PR #685, opened 2026-03-25
Fix indicator not opening when the translucent style is active

leolost2605:
Currently indicator can't be opened when the panel is in the translucent style. This fixes that.
I don't really know why this breaks it but I also don't seem to see any difference without this so I decided to just remove it.

### DIFF
--- a/data/styles/Application.css
+++ b/data/styles/Application.css
@@ -16,7 +16,6 @@ panel.maximized {
 
 panel.translucent > box {
     border-radius: 5px 5px 0 0;
-    margin-bottom: 4px;
 }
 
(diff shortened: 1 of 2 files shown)


danirabbit:
This margin is here because of the box shadow. Without it, the box shadow gets clipped. I think now we can maybe use 'overflow = VISIBLE`? But I was looking at this the other day and the dock does some other stuff to make sure struts are set correctly while not clipping shadows

leolost2605:
Ahh ok I added some manual size calculation to make sure we include size for the box shadow without it messing with something.

danirabbit (approved):
Nice! Thanks :)

PR #695, opened 2026-08-26
Application: don't manually register session

danirabbit:
This was replaced with a application property but according to docs Gtk always registers with the session manager these days: https://valadoc.org/gtk4/Gtk.Application.register_session.html


meson.build
```diff
@@ -29,7 +29,7 @@ gio_dep = dependency('gio-2.0')
 gio_unix_dep = dependency('gio-unix-2.0')
 gmodule_dep = dependency('gmodule-2.0')
 gdk_wl_dep = dependency('gtk4-wayland')
-gtk_dep = dependency('gtk4')
+gtk_dep = dependency('gtk4', version: '>=4.22')
```
leolost2605: Since it still works perfectly fine without this can we drop this?

leolost2605 (approved):
I don't know too much about this but from what I found:

So it seems that since https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8769 (which is the same MR that deprecated register_session) this doesn't actually register anything anymore. However since wingpanel isn't a required component in the session settings ig this is fine anyways?

PR #696, opened 2026-08-26
Move sort func to IndicatorEntry

danirabbit:
This is only called from one place. Move the compare function to the class it compares, as has been the recent code style


src/Widgets/IndicatorEntry.vala
```diff
@@ -35,6 +35,12 @@ public class Wingpanel.Widgets.IndicatorEntry : Granite.Bin {
         }
     }
 
+    private const string UNKNOWN_INDICATOR = "xxx-unknown";
+    private const string AYATANA_INDICATOR = "xxx-ayatana";
```
leolost2605: I think we usually have constants at the top so maybe keep that for consistency

leolost2605 (approved):
One nit apart from that LGTM

PR #697, opened 2026-08-26
IndicatorEntry: misc cleanup

danirabbit:
* Monitor indicator close and visible in the popover manager. This lets us be more accurate/concise with some actions
* Make revealer transition setting internal and private
* Get rid of an extra `halign`
* Early return on indicator bar being null
* Don't need to hold class refs to controllers manually
* Bind revealer to visible property


src/Services/PopoverManager.vala
```diff
@@ -129,5 +126,17 @@ public class Wingpanel.Services.PopoverManager : Object {
...
+
+        widg.base_indicator.notify["visible"].connect (() => {
+            /* order will be changed so close all open popovers */
+            popover.popdown ();
+
+            if (!widg.base_indicator.visible && get_visible (widg)) {
+                current_indicator = null;
+                widg.display_widget.has_tooltip = true;
+            }
+        });
```
lenemter, line 140: This looks more complicated than before. Why is this necessary?

danirabbit: I can't remember what I was doing here. Gonna draft these and break up individual changes

danirabbit:
Closing because #174 is a way better direction

PR #699, opened 2026-08-27
Open Notifications with keyboard shortcut

### LINKED ISSUE #313: Open with a keyboard shortcut
### Problem

There's currently no quick way to access the notifications list via keyboard when a new notification comes in

### Proposal
- [x] Open with notifications indicator with a keyboard shortcut.
- [x] Propose a branch to shortcuts overlay to list this as a system shortcut (could also just file a feature request and someone else can pick it up)

 Super + N makes sense to me and doesn't seem to be assigned to anything. Though there's some prior art (GNOME) for that jumping to the active notification instead of notification center

### Prior Art (Optional)

Super + N seems to be what Windows uses: https://www.windowscentral.com/windows-11-four-new-keyboard-shortcuts

Super + V was suggested here as it's the default in GNOME: https://github.com/elementary/notifications/issues/238

Most sources I could find said there is no shortcut by default in macOS, but [this source](https://mefmobile.org/how-to-invoke-a-specific-notification-center-view-using-keyboard-shortcuts-on-mac/) said Alt + F1

danirabbit:
fixes https://github.com/elementary/panel-notifications/issues/313

### DIFF
--- a/schemas/io.elementary.desktop.wingpanel.gschema.xml
+++ b/schemas/io.elementary.desktop.wingpanel.gschema.xml
@@ -1,15 +1,22 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <schemalist>
-	<schema path="/io/elementary/desktop/wingpanel/" id="io.elementary.desktop.wingpanel" gettext-domain="wingpanel">
-		<key type="b" name="use-transparency">
-			<default>true</default>
-			<summary>Sets if the panel uses transparency.</summary>
-			<description>Disable this to provide higher contrasts and make indicators better readable.</description>
-		</key>
-		<key type="b" name="scroll-to-switch-workspaces">
-			<default>false</default>
-			<summary>Sets if scrolling on the panel switches workspaces.</summary>
-			<description>Sets if scrolling on the panel switches workspaces.</description>
-		</key>
-	</schema>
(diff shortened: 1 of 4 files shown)


lenemter:
I'm against doing this in the wingpanel itself and I believe indicators should handle this themselves

danirabbit:
@lenemter I don't think indicators can open themselves?

lenemter:
Then we should either add API for that, or just use existing undocumented toggle-indicator action

danirabbit:
@lenemter tbh I don't think it's a good idea for indicators to be able to open themselves. I think the panel should be in charge of that

lenemter:
I do agree with you here. I just don't like that wingpanel essentially hardcodes shortcuts just for the notification indicator

The only solution I can think of right now is to provide indicator API for registering these shortcuts

danirabbit:
@lenemter can we make that a task to follow up? I'd like to just ship this feature and I know I probably won't get around to adding a new accels API anytime soon 😅

lenemter:
@danirabbit OK. I can implement this API once I get OS 9 install, but please add a TODO comment and create a new issue so we won't forget about this

lenemter (approved):

danirabbit:
@lenemter appreciate you!

PR #701, opened 2026-08-27
BackgroundManager: Defer to user preference when translucent

### LINKED ISSUE #657: Wingpanel theming
### Problem

The wingpanel seems to be dark/light/semi-transparent depending on the kind of desktop background that is being used. Regardless of whether or not the system theme is set to light or dark. The panel seems to theme itself.

It would be nice to have the option to modify these settings individually. 

See attached screenshots:

Wingpanel (fully-transparent)

Image

Wingpanel (semi-transparent dark)

Image

Wingpanel (semi-transparent light)

Image

### Proposal

Allow the user to choose whether or not they would like the wingpanel to be fully transparent or semi-transparent, as well as keeping the color of the wingpanel consistant with the system theme. (i.e dark/light) instead of having the system automatically determine this based on wallpaper contrast. 

Thanks!

### Prior Art (Optional)

_No response_

danirabbit:
fixes #657

Follow the user preference when the panel will be translucent.

Screenshot from 2026-08-27 17 41 01
Screenshot from 2026-08-27 17 41 29

If Gala thinks the panel should be transparent, reduce transparency makes it translucent:

Screenshot from 2026-08-27 17 45 30
Screenshot from 2026-08-27 17 45 42

if Gala thinks the panel should be translucent, reduce transparency makes it completely opaque

Screenshot from 2026-08-27 17 40 13

Screenshot from 2026-08-27 17 41 14

If Gala thinks the panel should be transparent, maintain the current behavior:

Screenshot from 2026-08-27 17 41 54
Screenshot from 2026-08-27 17 42 06

Note: This should theoretically use Granite.StyleManager, but it's not sending correct color info rn

### DIFF
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -21,7 +21,6 @@ public class Wingpanel.PanelWindow : Gtk.Window {
     public Services.PopoverManager popover_manager;
 
     private Widgets.Panel panel;
-    private int panel_height;
 
     private Pantheon.Desktop.Shell? desktop_shell;
     private Pantheon.Desktop.Panel? desktop_panel;
@@ -75,7 +74,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
         ((Gdk.Toplevel) get_surface ()).compute_size.connect (on_compute_size);
 
         update_panel_dimensions ();
-        Services.BackgroundManager.initialize (panel_height);
+        Services.BackgroundManager.initialize (panel.get_height ().clamp (16, 48));
 
         init_wl ();
     }
@@ -87,8 +86,6 @@ public class Wingpanel.PanelWindow : Gtk.Window {
     }
 
     private void update_panel_dimensions () {
-        panel_height = panel.get_height ();
-
         // We just use our monitor because Gala makes sure we are always on the primary one
         var monitor_dimensions = get_display ().get_monitor_at_surface (get_surface ()).get_geometry ();
 
--- a/src/Services/BackgroundManager.vala
+++ b/src/Services/BackgroundManager.vala
@@ -132,20 +132,20 @@ namespace Wingpanel.Services {
 
             bus.state_changed.connect ((state, animation_duration) => {
                 current_state = state;
-                state_updated (state, animation_duration);
+                state_updated (animation_duration);
             });
 
             state_updated ();
             return true;
         }
 
-        private void state_updated (BackgroundState state = current_state, uint animation_duration = 0) {
+        private void state_updated (uint animation_duration = 0) {
             if (!use_transparency) {
                 background_state_changed (BackgroundState.MAXIMIZED, animation_duration);
                 return;
             }
 
-            switch (state) {
+            switch (current_state) {
                 case TRANSLUCENT_DARK:
                 case TRANSLUCENT_LIGHT:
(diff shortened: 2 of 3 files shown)


src/PanelWindow.vala
```diff
@@ -75,7 +74,7 @@ public class Wingpanel.PanelWindow : Gtk.Window {
         ((Gdk.Toplevel) get_surface ()).compute_size.connect (on_compute_size);
 
         update_panel_dimensions ();
-        Services.BackgroundManager.initialize (panel_height);
+        Services.BackgroundManager.initialize (panel.get_height ().clamp (16, 48));
```
danirabbit: I guess this is a workaround, but it fixes the transparent behavior 🤷‍♀️

wpkelso: Is it worth writing a note that this is considered a workaround, rather than "ideal" behavior?

danirabbit: Yeah hoping for some insight maybe from @lenemter

amz-x: Would the following not suffice?
```vala
update_panel_dimensions ();

int minimum_height, natural_height;
panel.measure (Gtk.Orientation.VERTICAL, -1, out minimum_height, out natural_height, null, null);
Services.BackgroundManager.initialize (natural_height);

init_wl ();
```

src/Services/BackgroundManager.vala
```diff
@@ -53,6 +54,7 @@ namespace Wingpanel.Services {
             }
         }
 
+        private Gtk.Settings style_manager;
```
lenemter: Why is it named `style_manager`? It can easily be confused with Granite.StyleManager

src/Services/BackgroundManager.vala
```diff
@@ -137,8 +142,41 @@ namespace Wingpanel.Services {
...
+                            background_state_changed (TRANSLUCENT_DARK, animation_duration);
+                        }
+                        return;
+                    default:
+                        background_state_changed (BackgroundState.MAXIMIZED, animation_duration);
+                        break;
+                }
+
+                return;
+            }
```
lenemter, line 147: Why do we use translucent styles here? They are transparent

lenemter (changes requested):

src/Services/BackgroundManager.vala
danirabbit: It *should* use style manager but doesn't currently. I'll rename to gtk_settings for now

src/Services/BackgroundManager.vala
danirabbit: My idea was to just increase contrast instead of going fully opaque here. If gala thinks there's enough contrast to be fully transparent that a semi-transparent style should be very contrasty

lenemter: Makes sense but can you add a comment explaining that please?

danirabbit: I guess it's not quite AAA here. I can remove this and revisit it later when we have better contrast media queries

Image

lenemter (changes requested):
Also, Gtk.Settings is a static object, no need to store it in a private field

lenemter (approved):
LGTM

PR #703, opened 2026-08-28
Popover: remove arrow, rounder, margins

danirabbit:
Screenshot from 2026-08-27 18 07 18
Screenshot from 2026-08-27 18 07 30

### DIFF
--- a/data/styles/Application.css
+++ b/data/styles/Application.css
@@ -108,7 +108,12 @@ panel.translucent.color-light > box {
     -gtk-icon-shadow: none;
 }
 
+popover.indicator {
+    margin: 0.5rem;
+}
+
 popover.indicator > contents {
+    border-radius: 0.75rem;
     min-width: 20rem;
     padding: 0.25rem 0;
 }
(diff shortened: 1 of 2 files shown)


kurisubrooks:
I think my main point of contention is that if the arrow is removed, we need to signify that this popover is connected to X icon/panel item in another way. 

Currently there's no active state for wingpanel items when a popover is open, so it just looks like another floating window with no relation to the content it opens from. 

I also can't remember off the top of my head if we currently animate the popover contents, but adding a fade in + down animation to the popover would also help anchor the content to the panel item were the arrow removed.

kurisubrooks:
Depends on #704

danirabbit:
@kurisubrooks animating the popover will take some more time, but how is it now with the checked state?

kurisubrooks:
> @kurisubrooks animating the popover will take some more time, but how is it now with the checked state?

I think the animation can be handled in a follow-up. With the :checked state, it should be sufficient for now. Did you want me to spin it off into an issue?

kurisubrooks (approved):
I'll create a follow-up issue for the animation, so this isn't blocked. Thanks 👍🏻 

Edit: tracked in #705

PR #706, opened 2026-08-29
Rework size request and layouting

### LINKED ISSUE #693: Wingpanel width not updating upon display resolution change, elementary OS 9 Daily
### What Happened?

Changing the display resolution whilst logged in does not change the width of wingpanel, it only changes upon logging back in, however changing the display scaling does make wing panel update correctly as intended

this may potentially be the same as issue #689 but i have not been able to reproduce this on stable 

### Steps to Reproduce

1. Open display settings
2. change the resolution and/or the screen orientation
3. wingpanel does not change its width as expected

4. change the display scaling
5. wingpanel does adjust as expected

### Expected Behavior

wingpanel should adjust itself as expected as in older versions or as when changing the scaling

### OS Version

Early Access

### OS Architecture

arm64 (on Raspberry Pi, Apple Silicon, etc.), amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Early Access or Compiled from git

### Log Output

```shell

```

### Hardware Info

tested on elementaryos-9.0-daily-arm64.20260815.iso through UTM on macOS and elementaryos-9.0-daily-amd64.20260815.iso on gnome boxes

### LINKED ISSUE #702: Blur region is incorrect
### What Happened?

The blurred region of the panel stops before the end of the panel

### Steps to Reproduce

Have a panel with a translucent background. This is especially easy to see with really busy wallpapers:

Image
Image

### Expected Behavior

blur region should match the height of the panel minus the shadow area

### OS Version

Early Access

### OS Architecture

amd64 (on most hardwares)

### Session Type

Secure Session (Wayland, This is the default)

### Software Version

Early Access or Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

lenemter:
Fixes https://github.com/elementary/wingpanel/issues/702
Fixes https://github.com/elementary/wingpanel/issues/693 (?)

Follows the method of requesting size we use in Dock, which proved to be robust

### DIFF
--- a/src/PanelWindow.vala
+++ b/src/PanelWindow.vala
@@ -18,13 +18,21 @@
  */
 
 public class Wingpanel.PanelWindow : Gtk.Window {
+    private class BottomMargin : Gtk.Widget {
+        public const int SIZE = 64;
+
+        construct {
+            height_request = SIZE;
+        }
+    }
+
     public Services.PopoverManager popover_manager;
 
     private Widgets.Panel panel;
-    private int panel_height;
 
     private Pantheon.Desktop.Shell? desktop_shell;
     private Pantheon.Desktop.Panel? desktop_panel;
+    private bool initialized_background_manager = false;
 
     private Gtk.CssProvider? style_provider = null;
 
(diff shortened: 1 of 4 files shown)


src/PanelWindow.vala
```diff
@@ -18,29 +18,43 @@
  */
 
 public class Wingpanel.PanelWindow : Gtk.Window {
+    private class BottomMargin : Gtk.Widget {
+        public const int SIZE = 64;
```
danirabbit, line 22: Is there any reasoning behind this particular number?

danirabbit (approved):
Nice! Works for me

src/PanelWindow.vala
lenemter: I just copied it from the Dock. It can be a lot lower (like 20), but it shouldn't affect anything

PR #708, opened 2026-08-29
Fix Fedora/OpenSUSE CI

lenemter:


.github/workflows/main.yml
```diff
@@ -42,7 +42,7 @@ jobs:
       - uses: actions/checkout@v7
       - name: Install Dependencies
         run: |
-          dnf install -y gala-devel libgee-devel gobject-introspection-devel glib2-devel granite-devel gtk3-devel meson valac
+          dnf install -y gala-devel libgee-devel gobject-introspection-devel glib2-devel granite-7-devel gtk3-devel meson valac
```
danirabbit: Still has gtk3-devel?

danirabbit (approved):

.github/workflows/main.yml
lenemter: Removed

PR #712, opened 2026-08-30
Panel: use Gtk.DirectionType instead of true/false

danirabbit:
Just a personal pet peeve


jeremypw (approved):
Makes sense, no regressions.

I note that indicator cycling action does not work reliably anyway but I guess this will be fixed at some point and it will be made more discoverable.

PR #714, opened 2026-09-01
Use ListModels / FlowBoxes for showing indicators

lenemter:
Changes:
- Instead of using Gtk.Box and manually adding/removing indicators from it, use Gtk.FlowBox + ListModel
- IndicatorBar now just shows the indicators and doesn't handle insertion/removal
- IndicatorEntry now doesn't remove itself from the IndicatorBar and just shows the indicator
- Panel now handles adding/removing/filtering/sorting all the indicators via ListModels

### DIFF
--- a/src/Widgets/IndicatorBar.vala
+++ b/src/Widgets/IndicatorBar.vala
@@ -17,45 +17,28 @@
  * Boston, MA 02110-1301 USA.
  */
 
-public class Wingpanel.Widgets.IndicatorBar : Gtk.Box {
-    private Gee.List<IndicatorEntry> sorted_items;
+public class Wingpanel.Widgets.IndicatorBar : Granite.Bin {
+    public ListModel indicator_entries { private get; construct; }
+
+    private Gtk.FlowBox flow_box;
+
+    public IndicatorBar (ListModel indicator_entries) {
+        Object (indicator_entries: indicator_entries);
+    }
 
     construct {
-        sorted_items = new Gee.ArrayList<IndicatorEntry> ();
+        flow_box = new Gtk.FlowBox () {
+            orientation = HORIZONTAL,
+            selection_mode = NONE
+        };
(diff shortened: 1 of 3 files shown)


danirabbit (changes requested):
Generally like this change, but we've broken revealer animations. Need to add a little switch statement in IndicatorEntry to set the correct reveal animation

danirabbit (changes requested):
Almost there. It looks like we're not getting the animate in? You can easily reproduce by setting the NightLight schedule to the opposite of sunrise/sunset and then toggling between the two

lenemter:
@danirabbit nightlight indicator seems to be broken for me, I don't get an icon, but the indicator is clickable. This happens with both this branch and main branch.

danirabbit:
@lenemter oh are you missing an icon update or something? It's working here:

[Kooha-2026-09-01-11-43-13.webm](https://github.com/user-attachments/assets/b6eaa386-d1ff-469f-a4d0-03fc6af05842)

lenemter:
It looks like this for me. I'm still on OS 8 but compiled Gtk 4.23.3, maybe this has something to do with it..?

image

danirabbit:
@lenemter for you :) https://github.com/elementary/panel-nightlight/pull/119

lenemter:
@danirabbit This is definitely an issue with the indicator, as it works for me with Gtk 4.22.0. I suppose the issue is that we're using Gtk.Spinner to show svg icon?

danirabbit:
[Kooha-2026-09-01-12-01-03.webm](https://github.com/user-attachments/assets/34343b41-ceed-459b-bdfd-f4fc43aa8209)


I can still reproduce even after updating to Gtk.SVG for the nightlight indicator

lenemter:
@danirabbit Should be fixed now

danirabbit (approved):

PR #715, opened 2026-09-01
Cleanup toggle-popover action

lenemter:


danirabbit (approved):
afaict this works and is way simpler

PR #716, opened 2026-09-01
Remove unused actions and command line options

lenemter:


danirabbit (approved):
Oh yeah that's interesting. We only use toggle 🤷‍♀️

PR #721, opened 2026-09-02
 Don't reuse the popover for all indicators

lenemter:
Not sure why we're doing this in the first place


danirabbit:
It was to avoid opening and close the popover during scrubbing. We'd just move the popover and change its contents. It would be nice to still have that behavior, which I what I was saying before about we should have one popover per IndicatorBar and then change `pointing_to` instead of reparenting. But at this point just fixing the bug would be nice :sweat_smile:

PR #724, opened 2026-09-06
BackgroundManager: avoid using deprecated GTK keys

danirabbit:


zeebok (approved):

zeebok:
Not sure why it is failing tests, I don't even see an error message?

danirabbit:
Because panel doesn't build for OS 8 anymore. Requires newer deps

PR #725, opened 2026-09-07
Store wingpanel height in settings schema

vjr:
Addresses https://github.com/elementary/gala/issues/2560

Blocks https://github.com/elementary/gala/pull/2924 to fix the issue.

### DIFF
--- a/schemas/io.elementary.desktop.wingpanel.gschema.xml
+++ b/schemas/io.elementary.desktop.wingpanel.gschema.xml
@@ -11,6 +11,11 @@
       <summary>Sets if scrolling on the panel switches workspaces.</summary>
       <description>Sets if scrolling on the panel switches workspaces.</description>
     </key>
+    <key type="i" name="height">
+      <default>-1</default>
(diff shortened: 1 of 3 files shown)


danirabbit (changes requested):
gsettings should not be used as IPC. There's dbus for that

vjr:
> gsettings should not be used as IPC. There's dbus for that

Not even for a single property/integer? See https://github.com/elementary/wingpanel/blob/main/src/Services/BackgroundManager.vala#L70 isn't this doing just that? Although, I'll switch to using DBus. Thanks.

danirabbit:
@vjr that's actually a setting from GTK settings so it's safe to use.

Glib.Settings are usually not safe to use because it will crash if keys are missing so you create a hard packaging dependency. And then we run into the issues of circular dependency. If the panel depends on Gala and Gala depends on the panel we get into a pickle. We need dependencies to go one way. So the rule of thumb is we shouldn't use other packages gsettings for settings if we can help it, and never for IPC

PR #727, opened 2026-09-12
Move ServerType enum to its own file

lenemter:
Breaks API compatibility. So all indicators have to be recompiled after this branch. This step is needed for https://github.com/elementary/wingpanel/pull/722 and right now is the best time to do it


ryonakano (approved):
Makes sense 👍

PR #728, opened 2026-09-12
Add layer shell support

stilic:
Still a bit broken but it does run fine under a few [River](https://isaacfreund.com/software/river/) window managers from my own testing.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -24,6 +24,8 @@ i18n = import('i18n')
 gnome = import('gnome')
 pkg = import('pkgconfig')
 
+gtkls_dep = dependency('gtk4-layer-shell-0')
+gala_dep = dependency('gala', version: '>= 8.3.0', required: false)
 glib_dep = dependency('glib-2.0', version: '>=2.32')
 gio_dep = dependency('gio-2.0')
 gio_unix_dep = dependency('gio-unix-2.0')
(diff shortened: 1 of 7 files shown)


stilic:
The big blocker is the broken keyboard focus: it won't move the focus to the popovers at all for now.

danirabbit:
@stilic thanks for this! Can you also make sure to add `libgtk4-layer-shell-dev` to the GitHub workflow for CI?

stilic:
> @stilic thanks for this! Can you also make sure to add `libgtk4-layer-shell-dev` to the GitHub workflow for CI?

Will do later today!

src/Services/PopoverManager.vala
```diff
@@ -57,17 +58,31 @@ public class Wingpanel.Services.PopoverManager : Object {
...
                 ((Widgets.IndicatorEntry)popover.parent).display_widget.has_tooltip = true;
+                GtkLayerShell.set_keyboard_mode (owner, GtkLayerShell.KeyboardMode.NONE);
                 popover.popdown ();
             }
         }
     }
 
-    construct {
+    public PopoverManager (Wingpanel.PanelWindow? owner) {
+        this.owner = owner;
```
garaevdi: This should use GObject style constructor i.e. something like
```vala
public PopoverManager (Wingpanel.PanelWindow? owner) {
    Object (owner: owner);
}

construct {
  // construct code
}
```

meson.build
```diff
@@ -24,6 +24,8 @@ i18n = import('i18n')
 gnome = import('gnome')
 pkg = import('pkgconfig')
 
+gtkls_dep = dependency('gtk4-layer-shell-0')
+gala_dep = dependency('gala', version: '>= 8.3.0', required: false)
```
garaevdi: Could this be under compile flag so there wouldn't be unused dependency at build time when compiling with gala?
```meson
use_gala = get_option('use_gala')
if use_gala
  gala_dep = dependency('gala')
  # other setup
else
  gtkls_dep = dependency('gtk4-layer-shell-0')
endif
```

src/Widgets/Panel.vala
```diff
@@ -151,8 +151,10 @@ public class Wingpanel.Widgets.Panel : Granite.Bin {
     private void begin_drag (double x, double y) {
         popover_manager.current_indicator = null;
 
+        #if USE_GALA
         var background_manager = Services.BackgroundManager.get_default ();
         background_manager.begin_grab_focused_window ((int) x, (int) y);
+        #endif
```
ryonakano: Pre-process directives should have no indentation to distinguish with normal steps. The same comment applies to all occurrences of `#if USE_GALA`

```vala
#if USE_GALA
        var background_manager = Services.BackgroundManager.get_default ();
        background_manager.begin_grab_focused_window ((int) x, (int) y);
#endif
```

meson.build
stilic: Do you think it's better if I add two flags for each instead of one?

PR #729, opened 2026-09-14
Fix scrubbing

lenemter:
Closes https://github.com/elementary/wingpanel/pull/721

Changes:
- Use `pointing_to` instead of reparanting the popover
- Remove unintuitive behaviour where setting current_indicator to the opened indicator toggled it


src/Services/PopoverManager.vala
```diff
@@ -34,50 +34,53 @@ public class Wingpanel.Services.PopoverManager : Object {
...
+                indicator_open = true;
             } else if (value == null && _current_indicator != null) { // Close requested
-                indicator_open = false;
-                _current_indicator.base_indicator.closed ();
-                _current_indicator.set_state_flags (NORMAL, true);
-                _current_indicator = null;
-            } else if (_current_indicator.base_indicator.code_name == value.base_indicator.code_name) { // Close due to toggle
+
                 _current_indicator.set_state_flags (NORMAL, true);
+                _current_indicator.display_widget.has_tooltip = false;
```
danirabbit: has_tooltip should be the opposite. If it's open, we hide the tooltip. If it's closed we show the tooltip

danirabbit (commented):
For some reason with the branch the only indicator that will open now is notifications

PR #732, opened 2026-09-18
Replace indicator order HashMap with a switch

lenemter:
Requires https://github.com/elementary/wingpanel/pull/731


danirabbit (approved):
So much simpler

PR #733, opened 2026-09-19
Simplify PopoverManager.current_indicator

lenemter:
Remove unintuitive behaviour where setting current_indicator to the opened indicator toggled it


danirabbit (approved):
This makes sense and works correctly as far as I can tell. I agree the previous behavior is weird and this makes it more clear

